PR #3, opened 2015-06-05
Added documentation for Serial batch submission for R

mikecroucher:


software/apps/R.rst
```diff
@@ -27,8 +27,36 @@ R can then be run with ::
 
         $ R
 
-Installing
-----------
+Serial (one CPU) Batch usage
+----------------------------
+Here, we assume that you wish to run the program :code:`my_code.R` on the system.
```
Cadair: FYI, I think you can just use double back ticks, rather than `:code:` as a rst shortcut

mikecroucher: Thanks!

PR #67, opened 2015-07-09
R updates

mikecroucher:


merge_upstream.sh
```diff
@@ -0,0 +1,3 @@
+git fetch upstream
+git merge upstream/master
```
Cadair: what is your reasoning behind this over `git pull upstream master`??

mikecroucher: None at all. This shouldn't have ended up being committed - was a mistake.
I learned about `git pull upstream master` long after I learned how to do it using the two commands above. By this time, I was set in my ways which included putting them in a two line script to save typing.

Cadair: ah fair enough!

PR #72, opened 2015-07-23
Sun Grid Engine stub

mikecroucher:
Added a section on Sun Grid Engine in the software section. Populated it with a recent FAQ. 

This is just a stub for now, we can populate with full information over time.

### DIFF
--- a/software/SunGridEngine.rst
+++ b/software/SunGridEngine.rst
@@ -1,0 +1,19 @@
+Sun Grid Engine
+===============
+
+
+Frequently Asked SGE questions
+------------------------------
+**How do you ensure that a job starts after a specified time?**
+
+Add the following line in your submission script ::
+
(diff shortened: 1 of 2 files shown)


Cadair:
I feel as though the SGE stuff should be separate from software. SGE is to my mind more a feature of the cluster for most users rather than some software running on it.

This particular snippet should probably go in this table: https://github.com/rcgsheffield/iceberg_software/blob/master/using-iceberg/sge-intro.rst#all-scheduler-options

mikecroucher:
I don't mind either way so happy to do as you propose. We are, however, going to have to fix navigation! I couldn't find that table when navigating from the front page.

Cadair:
yes, I agree. It's something I have been meaning to do for a while. Please feel free to have a crack at it.

PR #75, opened 2015-07-24
added instructions to install additional packages in R

philmikejones:
Based on discussion with Mike I've written instructions to install additional user packages in R to a personal package library.

### DIFF
--- a/software/apps/r.rst
+++ b/software/apps/r.rst
@@ -56,6 +56,33 @@ Graphical output
 ----------------
 By default, graphical output from batch jobs is sent to a file called :code:`Rplots.pdf`
 
+Installing additional packages
+------------------------------
+As you will not have permissions to install packages to the default folder, additional R packages can be installed to your home folder :code:`~/`. To create the appropriate folder, install your first package in R in interactive mode:
+
+Load R in interactive mode as described above. Install a package with:
+
+        install.packages()
+        
+You will be prompted to create a personal package library. Choose yes. The package will download and install from a CRAN mirror (you may be asked to select a nearby mirror, which you can do simply by entering the number of your preferred mirror).
+
+Once the chosen package has been installed, additional packages can be installed either in the same way, or by creating a .R script. An example script might look like:
+
+        install.packages("dplyr")
+        install.packages("devtools")
+        
+Call this using :code:`source()`. For example if your script is called :code:`packages.R` and is stored in your home folder, source this from an interactive R session with:
+
+        source("~/packages.R")
+        
+These additional packages will be installed without prompting to your personal package library.
+
+To check your packages are up to date, and update them if necessary, run the following line from an R interactive session:
+
+        update.packages(lib.loc = "~/R/x86_64-unknown-linux-gnu-library/3.2/")
+
+The folder name after ~/R/ will likely change, but this can be completed with tab autocompletion from the R session. Ensure :code:`lib.loc` folder is specified, or R will attempt to update the wrong library.
+
 R Packages that require external libraries
 ------------------------------------------
 Some R packages require external libraries to be installed before you can install and use them. Since there are so many, we only install those libraries that have been explicitly requested by users of the system.


software/apps/r.rst
```diff
@@ -56,6 +56,33 @@ Graphical output
 ----------------
 By default, graphical output from batch jobs is sent to a file called :code:`Rplots.pdf`
 
+Installing additional packages
+------------------------------
+As you will not have permissions to install packages to the default folder, additional R packages can be installed to your home folder :code:`~/`. To create the appropriate folder, install your first package in R in interactive mode:
```
Cadair: it might be worth recommending people install into `/data` instead. The quota is higher and the backups less frequent.

mikecroucher: I don't think we need to worry about quota too much. Back in the dark days, when we didn't have modern R, users were installing R and ALL packages in their home directory!!

I guess we could suggest that they install to `/data` but thats extra complexity and I'm not a fan of complexity.

Instead, I think that the best thing we could do for our users is install a set of the most popular packages as part of the default install of R. We could start doing it as of the next release.

Cadair: does R default to home? or is it set by the user?

philmikejones: That sounds great (having packages installed by default). Can these be kept regularly updated if users have no control over updating them?

In terms of this pull request, shall I leave it as is to merge in to the documentation? Or edit it to specify a different path (e.g. /data/R). Or something else?

mikecroucher: It's R's default. User can change it.
If we change the default, we may cause pain.

Cadair: it's true, it was just a suggestion as I have filled up my home dir on iceberg with python installs...

mikecroucher: @philmikejones Users have no control over updating the ones we install *but* they can install more up to date versions in their personal library right?

For a managed system, I lean more towards stability rather than keeping things bleeding edge. What if you are in the middle of a parameter sweep...we update the central packages...half of your sweep uses foo version x.1 and the other half uses foo version x.2

If sysadmins did that to me, I'd be an angry bunny!

Cadair: This is the same problem that we were discussing about anaconda the other day...

I am generally in favour of leaving it to the user to manage their stack.

mikecroucher: What we **could** do, however is ensure that we keep R itself up to date! Each new version of R would get the latest packages available at the time of install. That provides a comprimise between stability and new-ness.

Cadair: @mikecroucher that would work, but where do we draw the line for installed packages?

philmikejones: I agree re: bleeding edge/stability so if a few versions of R are around in module the appropriate versions of the packages are presumably kept with them, too?

As a new point release of R is used (e.g. when R moves from 3.2.x > 3.3.0) the packages will be re-built with this version and the binaries will not necessarily be compatible with older versions of R.

I should have been clearer, but when new versions of R are added to modules can the packages be rebuilt for this version without affecting previous module versions?

Phew.

mikecroucher: @cadair we make it up as we go along, try to be sensible and allow the user to override whenever they like. Try to intelligently apply the 80/20 rule. I don't want to spend the rest of my natural life helping people install packages if I can get away with it.

@philmikejones Each version of R on iceberg would have its own set of packages. Even 3.2.x -> 3.2.y would have different packages if that's what we want.

mikecroucher:
This breaks the build but I suspect that's my fault. I'm going to merge, figure out what I've done wrong and fix.

software/apps/r.rst
mikecroucher: @philmikejones Congratulations on being the first Iceberg user to contribute to the documentation! Thank you too!

philmikejones: 3.2.x > 3.2.y not necessary because binaries unchanged.
Binaries only change between 3.x > 3.y

philmikejones: I think I stuffed up the code sections. I've not used this syntax before so missed ::
Will resubmit as new pull request

mikecroucher: ``3.2.x > 3.2.y not necessary because binaries unchanged.
Binaries only change between 3.x > 3.y``

Yes but because of the way we install R using environment modules, each installation will have a complete, independent set of packages. This helps our cause because it means that we can keep more up to date with packages. 

3.2.y will have later packages than 3.2.x. Each install is stable. The overall picture is up to date.
Or are you saying that packages *never* update between 0.0.x releases of R?

At worst, we duplicate packages.

PR #101, opened 2015-08-19
PGI Compiler 15.7 installed and documented

mikecroucher:


software/apps/anaconda.rst
```diff
@@ -58,6 +58,9 @@ Some modules were requested by users that were not included in the standard Anac
   module load apps/binapps/anacondapython/2.3
   conda install wxPython
 
+  module load mpi/gcc/openmpi/1.8.8
+  pip install mpi4py
```
Cadair: oh this is not a good idea.

It will be fine for now, but it would be better to tell the users to create conda envs for this stuff. mpi4py can make other modules, specifically h5py do weird things.

mikecroucher: OK no worries. I'll undo it and document mpi4py accordingly. It was
requested by a user so we need to do something.
Informing how to install it into their home directory is fine.

PR #144, opened 2015-10-20
move git to apps

Cadair:
@mikecroucher was there a reason for it to be in compliers when it was under apps in the modules dir?


mikecroucher:
git is a development tool and we agreed to change 'compilers' to 'development tools' in the documentation

https://github.com/rcgsheffield/iceberg_software/issues/109

It would be too much hassle to create a 'devtools' subdir for modules and install paths on the current system but we should perhaps do so for the new system?

PR #146, opened 2015-10-20
Deniz's changes from unify branch.

Cadair:


software/compilers/index.rst
```diff
@@ -1,7 +1,7 @@
-.. _compilers:
+.. _Development Tools:
```
Cadair: This should be one word (or hyphens) and by convention lowercase.

software/modules.rst
```diff
@@ -0,0 +1,61 @@
+.. _modules_usage:
+
+******************
+Modules on Iceberg
```
Cadair: To be consistent across all the files the first title in a page should use `====` as the underline, and should be underline only.

software/modules.rst
```diff
@@ -0,0 +1,61 @@
...
+******************
+
+ 
+
+In general the software available on iceberg is loaded and unloaded via the use
+of the modules system [#env-modules]_. 
+
+Modules make it easy for us to install many versions of different applications, compilers and libraries side by side and allow users to setup the computing environment to include exactly what they need. 
+
+**Important Note:** Modules are not available on the login node. You must move to a worker node using either ``qrsh`` or ``qsh`` before any of the following commands will work.
```
Cadair: It might be a good idea to use the 

```
.. note::
    Modules are not ....
```

Sphinx environment here.

Cadair: Also, you should interlink to getting started here for instructions on how to run qrsh etc.

Cadair:
@cs1ds I added some inline comments where I think some things should be changed. Nothing that important though.

software/index.rst
```diff
@@ -1,9 +1,9 @@
 .. _software:
 
-Iceberg Software
-================
+Software available on iceberg
```
Cadair: All the other titles are CamelCase.

Cadair: All the other titles are Title Case, so I think this one should be as well.

using-iceberg/index.rst
```diff
@@ -1,21 +1,19 @@
 Using iceberg
 =============
 
-The first step in using iceberg is getting an account and running an interactive
-session, if you don't know how to do this you should start with 
-:ref:`getting-started`.
+Having obtained an iceberg account, you can access iceberg from anywhere to either use it interactively or
```
Cadair: I do not like this change. I think that the new text is much less informative to a new user. The reason I put this sentance in at the start here is to make sure that people are entering this section with the prerequistite knowledge, i.e. an account and the ability to run an interactive job.

using-iceberg/index.rst
```diff
@@ -1,21 +1,19 @@
...
+to submit long-running batch jobs to its scheduling system ( SGE ). 
 
-Once you have connected to iceberg in interactive mode you can run jobs which 
-take less than **8 hours** and which are graphical or require user interaction.
-For more information on what applications you can run read :ref:`software`.
+SGE scheduler controls the way the cluster resources are shared by all the iceberg users. It gives each user the
```
Cadair: This sentance is good, but I don't feel it relaces the other sentances you have deleted. I think that having a long description of the different types of uses makes it easier for new users to discover what links they want to follow.

Cadair:
@cs1ds I added more comments on your new commits.

Cadair:
@cs1ds merged.

PR #201, opened 2015-12-02
Minor MATLAB fixes and updates

mikecroucher:


Cadair:
the html build seems to have updated properly.

PR #206, opened 2015-12-08
Add instructions for connecting to Jupyter Notebook server on Iceberg from own machine

willfurnass:
Rather convoluted, but a starting point.  

Might benefit from brief notes on `IPython.parallel`.


mikecroucher:
Thanks Will but as discussed offline, we have docs already for Jupyter integration.

Cadair:
I do wonder if it would be worth having a section on SSH tunnels in the docs somewhere, they are very useful for many things other than the Jupyter notebook?

PR #218, opened 2015-12-18
unified HDF5 documentation

Cadair:
ping @mikecroucher


mikecroucher:
It's a start but the whole HDF5 situation feels very messy to me. So. Many. Versions!

I'm also not a fan of the fact the the installation instructions are so brief. We should do better. I'll fix for the PGI version I built.

Cadair:
I have a build script for my latest version. It's already in the repo, need
to link it I think.

PR #221, opened 2015-12-22
Add an A-Z Software Index

Cadair:
screenshot from 2015-12-22 14-14-40


Cadair:
Unfortunately the limitation of this approach is that the index has to be seperate from the other documetation as it is built independantly by sphinx, hence the link out from the software/index page.

PR #223, opened 2015-12-22
Glossary

Cadair:
This converts the glossary to use the sphix glossary environment: http://sphinx-doc.org/markup/para.html?highlight=glossary#directive-glossary

You can then link to terms defined in the glossary by doing 
```
:term:`HPC`
```


Cadair:
ping @mikecroucher FYI

PR #258, opened 2016-02-11
Added runansys -fastdata flag description

susheel:
`runansys` command has a hidden flag `-fastdata` which allows the user to use /fastdata/$USER/$JOB_ID as the working directory


Cadair:
Thanks

PR #276, opened 2016-03-03
Scheduler command reference

mikecroucher:
The creation of this section was prompted by the development of our qrshx command which creates an X windows session that isn't disgusting (unlike qsh). We had nowhere to put these modifications and seem to have a few undocumented ones (Qsh for example which has been on Iceberg for ages but isn't documented anywhere)

I also wanted to start collecting a user-friendly reference section for scheduler commands that contained the most used options.


Cadair:
looking good!

PR #328, opened 2016-06-23
Added scheduler options following user query

mikecroucher:


using-iceberg/sge.rst
```diff
@@ -145,6 +145,11 @@ Command                Description
 
 -l mem=xxG             Specify the maximum amount (xx) of memory to be used.
 
+-l hostname=           Target a node by name. No recommended for normal use.
```
Cadair: 'Not recommened'

PR #334, opened 2016-07-11
Added Theano local build instructions

mondus:
Also has an update to README for windows users without make

### DIFF
--- a/gpu/Theano.rst
+++ b/gpu/Theano.rst
@@ -1,7 +1,7 @@
 .. _Theano:
 
 Deep Learning with Theano on GPU Nodes
------------------------
+--------------------------------------
 
 Theano is a Python library that allows you to define, optimize, and evaluate mathematical expressions involving multi-dimensional arrays efficiently. Theano is most commonly used to perform deep learning and has excellent GPU support and integration through PyCUDA. The following steps can be used to setup and configure Theano on your own profile.
 
--- a/gpu/Theano.rst
+++ b/gpu/Theano.rst
@@ -1,0 +1,41 @@
+.. _Theano:
+
+Deep Learning with Theano on GPU Nodes
+-----------------------
+
+Theano is a Python library that allows you to define, optimize, and evaluate mathematical expressions involving multi-dimensional arrays efficiently. Theano is most commonly used to perform deep learning and has excellent GPU support and integration through PyCUDA. The following steps can be used to setup and configure Theano on your own profile.
+
+Request an interactive session with a sufficient amount of memory:
+
+		qsh -l gpu=1 -P gpu --l gpu_arch=nvidia-k40m -l mem=13G
+
+Load the relevant modules
+
+		module load apps/python/anaconda3-2.5.0
+		module load libs/cuda/7.5.18
+
+Create a conda environment to load relevant modules on your local user account
+
+		conda create -n theano python=3.5 anaconda3-2.5.0 
+		source activate theano
+		
+Upgrade pip and install the other Python module dependencies which are required
+
+		pip install --upgrade pip
+		pip install theano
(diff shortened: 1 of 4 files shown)


gpu/Theano.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Create a conda environment to load relevant modules on your local user account
+
+		conda create -n theano python=3.5 anaconda3-2.5.0 
+		source activate theano
+		
+Upgrade pip and install the other Python module dependencies which are required
+
+		pip install --upgrade pip
+		pip install theano
```
Cadair: Is there a reason you are installing theano and these other packages from pip not conda? I notice there is no conda pycuda package.

gpu/Theano.rst
```diff
@@ -0,0 +1,41 @@
...
+		module load libs/cuda/7.5.18
+
+Create a conda environment to load relevant modules on your local user account
+
+		conda create -n theano python=3.5 anaconda3-2.5.0 
+		source activate theano
+		
+Upgrade pip and install the other Python module dependencies which are required
+
+		pip install --upgrade pip
```
Cadair: Doing this inside conda may not be a good idea. You are installing the pip (setuptools) package with conda then upgrading it with pip.

gpu/Theano.rst
```diff
@@ -0,0 +1,41 @@
...
+		qsh -l gpu=1 -P gpu --l gpu_arch=nvidia-k40m -l mem=13G
+
+Load the relevant modules
+
+		module load apps/python/anaconda3-2.5.0
+		module load libs/cuda/7.5.18
+
+Create a conda environment to load relevant modules on your local user account
+
+		conda create -n theano python=3.5 anaconda3-2.5.0
```
Cadair: do you need to install the whole anaconda stack rather than just the core dependencies of theano?

a quicker version of this command may be to clone the central anaconda3-2.5.0 env with: `conda create --clone anaconda3-2.5.0 -n theano`

Cadair:
Thanks a lot for this it looks great!

gpu/Theano.rst
mondus: You dont't *need* the whole of the anaconda stack, you could get away with just scipy and numpy but there are useful packages in there that many will want when using theano.

gpu/Theano.rst
mondus: Could be removed.

gpu/Theano.rst
mondus: No reason in particular other than I was previously using virtual-env with pip. conda-env now supports pip installed packages so I don't imagine it makes much difference which is used and dependencies are tracked in the environment from either install method.

Cadair: it doesn’t make a huge difference unless the pip package needs compiling. I am happy to leave it either way.

gpu/Theano.rst
Cadair: yeah makes sense.

gpu/Theano.rst
```diff
@@ -0,0 +1,41 @@
+.. _Theano:
+
+Deep Learning with Theano on GPU Nodes
+-----------------------
```
Cadair: Your underline is too short, it needs to match the length of the title.: https://travis-ci.org/rcgsheffield/iceberg_software/jobs/143932110#L970

mondus:
Have updated the branch with a fix for the RST underline issue

gpu/Theano.rst
Cadair: Can you remove this?

mondus: Removed the pip upgrade and commented that these could alternatively be installed with conda. I have not tested this however.

PR #337, opened 2016-07-21
update GPU docs to remove need to request a project

anthonybrookfield:
I've made a change to the scheduler so anyone can request GPUs.  Although the gpu project still exists in the scheduler config (to keep non-gpu from spilling onto the nodes) all users are automatically members of the project.  Also, the scheduler will automatically add the project request if you request GPUs to using "-P gpu" is no longer necessary.

I also rejigged things a bit to include batch info, rather than just interactive usage, to make it clear that batch usage is thing.

### DIFF
--- a/gpu/GPUAccess.rst
+++ b/gpu/GPUAccess.rst
@@ -1,7 +1,0 @@
-.. _GPUAccess:
-
-Requesting access to GPU facilities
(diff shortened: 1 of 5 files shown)


mikecroucher:
looks good to me.

PR #342, opened 2016-08-12
Added info on shared areas

unixnation:
Added additional info on how to access a shared area from Iceberg. This probably needs expanding a little more still.


anthonybrookfield:
That looks good.  Ta.

PR #384, opened 2016-10-13
MATLAB install instructions: revise on Iceberg; write for ShARC

willfurnass:


willfurnass:
Existing modulefiles contain
```
set     mcrroot  /usr/local/packages6/matlab/runtime/R2016a/v901
setenv MCRROOT $mcrroot
# below is recommended but directory does not exist so for now comment out
#  setenv XAPPLRESDIR $mcrroot//v901/X11/app-defaults
```
@cs1ds : Does the MCR need to be separately installed (and `MCRROOT` need to be set by our modulefiles)?  Do you know much about `XAPPLRESDIR`?  Looks like neither the `$mcrroot` nor `$matlabroot` for 2016a contain an `X11/app-defaults` subdirectory.

willfurnass:
@cs1ds?

willfurnass:
It appears that setting `$MCRROOT` makes it easier to use the ``mcc`` compiler and locate the MCR distributable.

PR #393, opened 2016-10-25
Remove run-time dependency of FFTW library on particular version of GCC

willfurnass:
At present the [`libs/gcc/5.2/fftw/3.3.4` modulefile has a run-time dependency on `compilers/gcc/5.2`](iceberg/software/libs/fftw.rst) when this is not necessary (as confirmed by running `ldd` on all binaries in the fftw installation directory).  This makes it more difficult to use fftw 3.3.4 with CUDA as the latter requires an older version of GCC.

The line `module load compilers/gcc/5.2` needs removing from the `libs/gcc/5.2/fftw/3.3.4` modulefile.

Also, I've made a few other refinements to the fftw modulefile:

* use `CPATH` instead of `CPLUS_INCLUDE_PATH` as fftw is a C library;
* set `MANPATH`
* set `PATH`

### DIFF
--- a/iceberg/software/libs/fftw.rst
+++ b/iceberg/software/libs/fftw.rst
@@ -62,6 +62,8 @@ Modulefile is on the system at ``/usr/local/modulefiles/libs/gcc/5.2/fftw/3.3.4`
 
   set FFTW_DIR /usr/local/packages6/libs/gcc/5.2/fftw/3.3.4
 
+  prepend-path CPATH $FFTW_DIR/include
   prepend-path LD_LIBRARY_PATH $FFTW_DIR/lib
-  prepend-path CPLUS_INCLUDE_PATH $FFTW_DIR/include
   prepend-path LIBRARY_PATH $FFTW_DIR/lib
+  prepend-path MANPATH $FFTW_DIR/share/man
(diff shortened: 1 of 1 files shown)


willfurnass:
@mikecroucher any comments re `CPATH` vs `CPLUS_INCLUDE_PATH `?

mikecroucher:
@willfurnass *shrugs*. 
Those environment variables cause me problems.

Some docs say you need to use `CPLUS_INCLUDE_PATH` for fftw.  Like this one https://github.com/dealias/fftwpp

Is it right? I dunno?!

I'm a little uncomfortable with tidying up an in-use module any more than is necessary. How do you know if it will break anything?

Removing the runtime dependency on gcc 5.2 is sensible. We have a good reason to do so.

The `CPATH` vs `CPLUS_INCLUDE_PATH` thing may well be fine....or it may not be. It clearly worked fine as it is so perhaps leave it?

Of course, if you wanted to use `CPATH` instead of `CPLUS_INCLUDE_PATH` for a newer version of fftw, I'd have no objections at all.

willfurnass:
Fair point; reverted the replacement of `CPLUS_INCLUDE_PATH` with `CPATH`.

PR #397, opened 2016-10-27
Reformat FLAME GPU info

willfurnass:
Convert [this page](parallel/gpu/FLAMEGPU.rst) from Markdown to reSTructuredText.

Also, make minor formatting changes to the [submitting GPU jobs](parallel/gpu/gpu_jobs.rst).

### DIFF
--- a/iceberg/software/apps/perl.rst.next
+++ b/iceberg/software/apps/perl.rst.next
@@ -1,0 +1,53 @@
+Perl
+====
+
+.. sidebar:: Perl
+
+   :Latest Version: 5.24.0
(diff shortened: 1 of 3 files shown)


willfurnass:
@mondus @mikecroucher  Need to think about where this info goes in the docs.  Do we want a guide for installing it?  Or also versions installed cluster-wide?  Either way, think this should go under [software](software).

mondus:
I agree. The Theano docs should also move to software once a module is created (me @mikecroucher and Twin have set aside some time doing doing this in Nov)

PR #413, opened 2016-10-31
ShARC: Intel PS 2017 install script and modulefiles

willfurnass:
Still need to:
 * Test this config
 * Update docs (inc. check links to scripts and modulefiles).


willfurnass:
Hmm, not activating during the install...

PR #414, opened 2016-10-31
NAG compiler v6.0 and v6.1: add docs for ShARC

willfurnass:
Plus testing example for ShARC and Iceberg.

Could also roll NAG compiler 6.1 docs + install script + modulefiles for ShARC and Iceberg into this PR when get a new license from NAG.


willfurnass:
Need to replace quoted modulefiles with links to ShARC and Iceberg modulefiles.

PR #417, opened 2016-11-01
ShARC OpenMPI + Intel compiler

willfurnass:
Docs, install script and modulefile for OpenMPI 2.0.1 built using the Intel 17.0.0 compiler.


willfurnass:
Add some text to better distinguish the Intel and GCC flavours of OpenMPI available?

PR #423, opened 2016-11-02
 added pgi nag library doc

cs1ds:


mikecroucher:
You can get rid of the $NAGINC environment variable by adding extra paths to the standard environment variables in the modulefile. See 

https://github.com/rcgsheffield/sheffield_hpc/blob/master/iceberg/software/modulefiles/libs/intel/15/NAG/fll6i25dcl 

for an example

cs1ds:
Thanks, I did try that and did not seem to work with the pgi compiler. It is needed to pick up the NAG module file.
I shall have another look if you managed to get it working with Intel.

mikecroucher:
No worries Deniz. There was nothing wrong with your modulefile.

The problem we with the install of the PGI Compiler. It was ignoring the CPATH environment variable.  These are gcc-specific variables that are standard enough that Intel respects them by default....PGI, not so much it seems.

Anyway, Ive tweaked the PGI install and now this works:

```
module load libs/pgi/15/NAG/fll6a24dpl 
module load compilers/pgi/15.10 
pgf90 a00aafe.f90 -lnag_nag  -o your_code.exe

 ./your_code.exe 

 A00AAF Example Program Results
 
 *** Start of NAG Library implementation details ***
 
 Implementation title: Linux/pgf90 64-bit
            Precision: FORTRAN double precision
         Product Code: FLL6A24DPL
                 Mark: 24.3 (self-contained)
 
 *** End of NAG Library implementation details *** 
```

willfurnass:
A couple of comments from someone who's yet to use the NAG libraries (and therefore may be asking naieve questions):
 * "*For many routines, you may find it beneficial to use the parallel version of the library.*" - do we provide these?  If not, how can they be accessed?
 * Why do we provide Mark 24 (presumably that means version 24) of the PGI version but Mark 25 of the Intel version?
 * Would the page be neater/clearer if the Intel and PGI sections were amalgamated (an open question)?
 * Would be good to include the modulefile for `fll6a24dpl` in the repo and link to it from this page.

mikecroucher:
@willfurnass - 

> "For many routines, you may find it beneficial to use the parallel version of the library." - do we provide these? If not, how can they be accessed?

Not yet but we should. I'll add to the ever growing list of things to do.

>Why do we provide Mark 24 (presumably that means version 24) of the PGI version but Mark 25 of the Intel version?

NAG stagger their releases. When a new version comes out (Mark 25 say), they release it for their most popular implementations first. Other implementations come out over time.

>Would be good to include the modulefile for fll6a24dpl in the repo and link to it from this page.

Yes

willfurnass:
Also, is the `libs/nag/24` modulefile now a legacy thing?

mikecroucher:
@willfurnass 

Probably although some may be using it so we should delete with care

willfurnass:
> NAG stagger their releases. When a new version comes out (Mark 25 say), they release it for their most popular implementations first. Other implementations come out over time.

Think it would be good to make a note of this in the docs.

cs1ds:
Merging for now so as to use it in training today.Two issues still outstanding but we will create new issues later.

PR #425, opened 2016-11-03
Iceberg: NAG Fortran compiler v6.1

willfurnass:
Inc. docs, install script and modulefile.

Also neatened up the docs and modulefile for ShARC.


mikecroucher:
What do `pushd` and `popd` do? Never seen them before.

willfurnass:
@mikecroucher `pushd` and `popd` are bash builtins for [appending and removing directories from the directory stack](https://www.safaribooksonline.com/library/view/learning-the-bash/1565923472/ch04s05.html).   Useful if you change directory for a bit then go back to where you were without having to explicitly state where that was.

PR #431, opened 2016-11-07
Iceberg + Intel compiler 15.0: stop modulefile clobbering env vars

willfurnass:
The `modulefiles/compilers/intel/15.0.3` module file was reassigning rather than prepending to `CPATH` and `INFOPATH`.

NB yet to be updated in production.


willfurnass:
Also added neater way of managing modulefile conflicts for the Intel compilers.

Issue #359, opened 2016-09-24
Install Intel Parallel Studio Composer Edition 2017


mikecroucher:
This is the first release of this series. It includes version 17 of the compilers.

I've put the installer in `/fastdata/fe1mpc/public`

willfurnass:
Doesn't contain tarballs; only RPMs and cabs.  Is this what's needed for adding a new version of the Intel compilers on our HPC machines?

willfurnass:
To do:

- [ ] Install script (see #413 for how done on ShARC)
- [ ] Module files for each of: compilers, TBB, IPP, MKL and DAAL  (see #413 for how done on ShARC)
- [ ] License config for compilers
- [ ] Testing
- [ ] Documentation for 5 components and PS as a whole  (see #413 for how done on ShARC)

NB Latest TBB is a priority as it's to be used for a forthcoming course (#412)

willfurnass:
Hmm, the Intel Parallel Studio Composer Edition installer may not like the OS on Iceberg:

```
$ iceberg/software/install_scripts/compilers/intel-ps-xe-ce/2017.0//install.sh
Missing optional prerequisites
-- Unsupported OS
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
There are some missing optional prerequisites. If you still want to continue,
please correct the CONTINUE_WITH_OPTIONAL_ERROR variable in silent configuration
file.
```

```
$ cat /etc/issue
Scientific Linux release 6.8 (Carbon)
```

PR #433, opened 2016-11-07
Iceberg + Intel MKL: add dedicated module file and page for v11.2.3

willfurnass:
Should make it easier to use GCC, apps/libs that depend on GCC for their C++ STL lib and the MKL in the same environment.

Version determined using [this useful table](https://software.intel.com/en-us/articles/which-version-of-the-intel-ipp-intel-mkl-and-intel-tbb-libraries-are-included-in-the-intel).

NB still need to install Intel Parallel Studio 2017 on Iceberg (#359)

### DIFF
--- a/iceberg/software/apps/intel_r.rst
+++ b/iceberg/software/apps/intel_r.rst
@@ -9,7 +9,7 @@ R (Intel Build)
    :URL: http://www.r-project.org/
    :Documentation: http://www.r-project.org/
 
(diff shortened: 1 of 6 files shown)


willfurnass:
@mikecroucher Any comments/suggestions?

mikecroucher:
Not at the moment. Looking good. That table is very useful indeed!

Thankfully, it seems that they've simplified the versioning as of 2017 release.

Issue #434, opened 2016-11-07
Install Relion 2 (beta)


willfurnass:
Relion 1.4 [already installed](http://docs.hpc.shef.ac.uk/en/latest/iceberg/software/apps/relion.html) on Iceberg.  Newer, CUDA-aware version requested by a research group.

Dependencies:

- [ ] OpenMPI
- [ ] CUDA 7.5
- [ ] [MotionCor2](http://msg.ucsf.edu/em/software/motioncor2.html) (single binary; also uses CUDA inc. CUFFT)
- [ ] CTFFIND3 ([already installed](http://docs.hpc.shef.ac.uk/en/latest/iceberg/software/apps/ctffind.html))
- [ ] [CTFFIND4](http://grigoriefflab.janelia.org/ctffind4)
- [ ] [GCTF](http://www.mrc-lmb.cam.ac.uk/kzhang/)
- [ ] RESMAP ([already installed](http://docs.hpc.shef.ac.uk/en/latest/iceberg/software/apps/resmap.html))
- [ ] FFTW (see #429 for new version that uses a CUDA-compatible GCC + libstdc++)
- [ ] FLTK ([already installed](http://docs.hpc.shef.ac.uk/en/latest/iceberg/software/libs/fltk.html) but does not use CUDA compatible GCC + libstdc++)

willfurnass:
@mikecroucher What's your take on modulefiles and docs for GCTF and MotionCor2 given that:

 **+** both are dependencies and [may need to be referenced in the module file](https://www2.mrc-lmb.cam.ac.uk/relion/index.php/Download_%26_install) for Relion 2 batch jobs _but_
 **-** both are *only* dependencies and are trivial to install?

mikecroucher:
My take is that its your call.  How likely are we to need them for anything else?

Can you install them both within the Relion2 directory? If so, perhaps just do that. They essentially won't exist other than for Relion2.

willfurnass:
@mikecroucher Good idea; done.

PR #435, opened 2016-11-07
Iceberg: add new version of CTFFIND (4.1.5)

willfurnass:
Depends on PRs #430 and #433.

An optional but requested dependency of RELION 2.0 (currently being installed; see #434)


willfurnass:
Travis will fail until PR #433 merged (and reStructuredText label `iceberg_intel_mkl` becomes defined).

PR #457, opened 2016-11-18
Fix anchor link padding

Cadair:
Have a small friday present, as I fixed this for another website where I definiantaly am not using the same css :D

### DIFF
--- a/themes/tuos/static/iceberg.css
+++ b/themes/tuos/static/iceberg.css
@@ -39,6 +39,14 @@ cite {
     background-color: #333132;
     border-color: #333132;
 }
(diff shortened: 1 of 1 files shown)


willfurnass:
Thanks; this gives us a little more space before each `div`, right?

Cadair:
yeah so that it doesn't vanish under the header bar.

willfurnass:
RTD, like Travis (see #474), is now failing to build the docs: https://travis-ci.org/rcgsheffield/sheffield_hpc/jobs/181412734  @Cadair any thoughts/suggestions re this (other than that it's definitely not CSS-related :))?

PR #507, opened 2016-12-15
Added test module files for Caffe and Torch

karmauk:


willfurnass:
* Looks like the Caffe version is wrong in the filesystem path or in the modulefile (should include `r3` or `rc3`)
* Was Caffe compiled on ShARC?  If so, the module file name and install path should have e.g. `/gcc-4.9.4` or `/gcc-4.9.4-cuda-8.0` after the Caffe version.  If not, then the Caffe version should be followed by `/binary`
* The Caffe modulefile doesn't appear to handle dependencies.  `ldd /usr/local/packages/libs/caffe/r3/bin/* | grep 'c++'` indicates that some binaries are dependent on a version of libstdc++ that is newer than the system version.  Having some `module load ...` or `module prereq ...` statements in your modulefile should help.

karmauk:
Made the module files also unload other incompatible modules e.g. CUDA 7. Is this acceptable behaviour?

willfurnass:
I think we could simplify the module pre-req checking logic for these modules and others if we use `module conflict some/path` for all compilers (GCC/Intel/NVCC):
- Include `module conflict dev/intel-compilers /dev/gcc` in the `dev/intel-compilers/*` module files (partly done)
- Include `module conflict dev/intel-compilers /dev/gcc` in the `dev/gcc/*` module files (not done)
- Include `module conflict `libs/CUDA` in the `libs/CUDA/*` module files (not done)

Also, should nccl be dependent on `libs/CUDA/8.0.44/binary` (for `libcudart.so.8.0`)?  If so, worth adding a `module load` to its modulefile and changing the path suffix to `binary-cuda-8.0.44` in case a CUDA 7.5 or 8.5 version is released.

Also, how important is the dependency on anaconda3?  I think Stuart (who set up Conda on ShARC) intended for the python packages provided by anaconda{2,3} modules to be regularly upgraded.  Also, forcing a module load ties people to Python 3.  Better to not load explicitly load a Python interpreter and say load anacondaX or conda (then create own conda env) if you want to use `pycaffe`?

karmauk:
I've made changes as you suggested. Please push it though for now as if we wait to fix everything it will be new years before I can even tell people to even start testing.

Right now Torch is using Atlas from a conda and Caffe uses boost, hdf5, python3.5 and numpy from conda. I know it seems like a bit of a hack at the moment but it was better to have something at least working now. I've moved the binaries in to its own /testing-<version> folder and changed the module file to reflect that.

Issue #512, opened 2016-12-19
Ensure that cannot load multiple compilers at once


willfurnass:
First discussed in #507.  Should add `module conflict some/path` for all compilers (GCC/Intel/PGI/NVCC):

- [ ] Include `module conflict dev/intel-compilers /dev/gcc` in the `dev/intel-compilers/*` module files
- [ ] Include `module conflict dev/intel-compilers /dev/gcc` in the `dev/gcc/*` module files
- [ ] Include `module conflict `libs/CUDA` in the `libs/CUDA/*` module files

PR #514, opened 2016-12-19
ShARC: Ensure multiple complier modules (GCC/Intel) or CUDA libraries cannot be loaded at the same time

willfurnass:
See issue #512


willfurnass:
Closing as ShARC to be decommissioned

Issue #489, opened 2016-12-12
Install Boost on ShARC


willfurnass:
Dependency of #488 

Use GCC 4.9.2

Dependencies:

- [x] icu
- [x] libunistring

PR #518, opened 2016-12-20
ShARC: Install script and docs for Boost library

willfurnass:
Requested in issue #489  

Note: **Do not merge** as install script **currently not working** as boost looks for and fails to find Python headers during compilation.  

Need to decide between the following:
 - build without python support
 - build after installing Python headers using an RPM?
 - build using headers provided by one conda environment (or a conda environment for each minor release of Python e.g. 2.7, 3.4 and 3.5)


mikecroucher:
Found myself here after talking to a user who is interested in Boost. Could we do option 2?

* build after installing Python headers using an RPM?

Looking at how I built Boost for Iceberg: 

http://docs.hpc.shef.ac.uk/en/latest/iceberg/software/libs/boost.html 

It seems that they 'Just Worked'. Could this be because the Python headers were already installed on Iceberg?   If so, this route should be fine since it's worked OK on Iceberg for a while now.

willfurnass:
Making boost dependent the system Python is not entirely desirable.  After learning that boost is available from the default conda channel: 
```
$ conda search boost
Fetching package metadata .........
boost                        1.57.0                        0  defaults        
                             1.57.0                        1  defaults        
                             1.57.0                        4  defaults        
                             1.60.0                   py27_0  defaults        
                             1.60.0                   py34_0  defaults        
                             1.60.0                   py35_0  defaults        
                             1.61.0                   py27_0  defaults        
                             1.61.0                   py34_0  defaults        
                             1.61.0                   py35_0  defaults        
                             1.61.0                   py36_0  defaults        
```
my preference would be to build and install boost without Python support and recommend that people use conda if they want boost + Python.  @mikecroucher Would this work in your case?

mikecroucher:
Probably. Let's give it a try

willfurnass:
- [ ] Install script
- [ ] Install
- [ ] Modulefile
- [ ] Documentation

PR #537, opened 2017-01-09
More info on working with SGE

willfurnass:
At present the 

Started drafting the following for iceberg:

 - Completed / aborted jobs
 - Deleting jobs
 - Queued and running jobs

**Work in progress.**

Want to duplicate for ShARC or just make minor edits to note how behaviour differs between iceberg and ShARC (e.g. `qacct` on ShARC == `qacct -u $USER` on iceberg)

Also need to add info on how accounting info is rotated as this has been recently enabled on ShARC and is to be enabled on Iceberg at some point soon.


mikecroucher:
This looks like good stuff but work on the branch seems to have stalled. Do you think that what's been done so far is good enough to be merged? Any work left over can be put in the Issues.

PR #559, opened 2017-02-10
docs and install info for tigervnc - used with virtualgl for remote vis

anthonybrookfield:
Blatant copy-pasta using the docs/install scripts for the Java package.


willfurnass:
We should decide on a way to combine this documentation with that added as part of PR #556 (which is on how to use qsh-vis on Iceberg).  My suggestion:
 - Motivation and general usage guide for both clusters (at this level: http://docs.hpc.shef.ac.uk/en/latest/hpc/index.html) acknowledging the differences between clusters. Subsections for TigerVNC approach and vglconnect/vglrun approach; list merits of both.  
 - Per-cluster guide to technical setup (modulefiles, installation notes) for both of TigerVNC and vglconnect/vglrun

anthonybrookfield:
closing - see #634

PR #576, opened 2017-03-13
Add specs for ShARC

willfurnass:
@mikeg64 @cs1ds I'd be grateful if you could check this (particularly the big mem node section).


mikecroucher:
Looks good to me. Is there a way we can confirm the specs via a script?

I'm merging this because I've been asked ShARC specs for a grant proposal and this is the best info we have.

willfurnass:
@mikecroucher Yes, auditing and producing a spec using a script would be vastly preferable - see #603.  The info here was pulled together for expediency.

PR #592, opened 2017-04-18
OpenSSL page for ShARC

willfurnass:
* Give explanation for installing OpenSSL inc. devel package as RPMs.
* Briefly explain risk of using conda-provided OpenSSL.


mikecroucher:
This looks great. Thanks Will

PR #616, opened 2017-05-04
Fix parameter for conda create and pip install

blodstone:


willfurnass:
@karmauk may be interested in this.

willfurnass:
Tensorflow docs have since been revised.  Closing.

PR #625, opened 2017-05-12
Singularity on ShARC: how installed and versioned.

willfurnass:
Note that not using module files and give reasons why.  @twinkarma  - any comments?


twinkarma:
Yes looks good to me.

PR #631, opened 2017-05-24
Fixed error in anaconda tensorflow environment creation code

twinkarma:


willfurnass:
+1 to not using a full anaconda install unless really needed :)

PR #635, opened 2017-05-31
Limit max version of Sphinx theme

willfurnass:
The `requirements.txt` file used by ReadTheDocs to build this Sphinx documentation previously asked that a version of `sphinx-bootstrap-theme` newer than 0.4.7 be installed.  It appears that following the release of 0.5.0 our ReadTheDocs-hosted documentation looks a bit odd.  I've therefore restricted the version of that package to >=0.4.7 but <0.5.0 until we've found a way of making the 0.5.0 styling and layout changes look a bit better.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,8 +39,8 @@ Building the documentation on a local Windows machine
 
 #. Create a new *conda environment* for building the documentation by running the following from this window: ::
 
-    conda create -n sheffield_hpc python=3.6 sphinx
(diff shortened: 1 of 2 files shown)


willfurnass:
Also updates `requirements.txt` to ensure that the version of Sphinx installed is the same used by default by ReadTheDocs - should ensure our dev envs and the build env on ReadTheDocs are more similar.

PR #655, opened 2017-06-20
ansys on sharc: rst, 2 * install and 2 * module files

dmrogers75:
Ansys 16.1 and 17.2 installed on Sharc. Both use the tight-integration mpi-rsh PE. One .rst file, two install scripts and two module files.


willfurnass:
Worth adding a comment under *Installation Notes* on what the `mpi-rsh` PE is and why it's needed?

dmrogers75:
Will do. There is a sentence in the "Batch jobs" which I will move and expand upon.

PR #665, opened 2017-07-04
updated notes/module/install scripts for OpenMPI 2.1.1 install

anthonybrookfield:
Install notes and files for OpenMPI 2.1.1. install - 

Currently installed for testing - trying to track down hard-to-reproduce issue where MPI processes don't use all the CPU cores they've been allocated by the scheduler.  I currently suspect a bug in OpenMPI - so try to reproduce using latest version.   (I'd be open to arguments for using the latest 2.0.x release instead if people think sticking on the same release is preferable....)

modulefile for install is currently only accessible to app-admins - will be made user-accessible if go live.

Note: the zip file containing the example media seems to be minor-version specific now(?) hence the change to the usual install script.

### DIFF
--- a/sharc/software/install_scripts/mpi/openmpi/2.1.1/gcc-6.2/install.sh
+++ b/sharc/software/install_scripts/mpi/openmpi/2.1.1/gcc-6.2/install.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+# This is a template script for building and installing software on ShARC.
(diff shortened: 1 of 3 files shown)


anthonybrookfield:
for testing renumbered module to be mpi/openmpi/0.2.1.1/gcc-6.2 and made world readable, cos restricting the permissions caused 'module' to generate warnings when run by normal users.

PR #671, opened 2017-07-10
Document SGE option '-l excl=true' for requesting exclusive access to node(s)

willfurnass:
Caveats also noted.  Addresses Issue #670.


anthonybrookfield:
Maybe add a comment to say that queuing time will typically be a lot longer if you use excl=true?

PR #674, opened 2017-07-12
Add note to warn users about loading external libs

annakrystalli:
I've added a note to warn users to load the external libraries r packages depend on before trying to install the packages themselves.


willfurnass:
Looks good.  Could also add a warning to Iceberg Intel R page, ShARC (normal) R page and ShARC Intel R page?

willfurnass:
@annakrystalli asked that this be closed as she wants to create a new conflicting PR.

PR #678, opened 2017-07-14
vasp on sharc: rst, install script and module files

dmrogers75:


sharc/software/apps/vasp.rst
```diff
@@ -0,0 +1,54 @@
...
+
+    #!/bin/bash
+    #$ -cwd
+    #$ -l h_rt=00:30:00
+    #$ -l rmem=2G
+    #$ -pe mpi 4
+
+    module load apps/vasp/5.4.1.05Feb16/intel-17.0.0-openmpi-2.0.1
+    
+    mpiexec -np $NSLOTS vasp_std
```
dmrogers75: Was using mpiexec from an example Vasp script - I will change to mpirun. So, should it be "mpirun -np 4 vasp_std"? Or just "mpirun vasp_std"?

sharc/software/apps/vasp.rst
```diff
@@ -0,0 +1,54 @@
...
+The script requests 4 cores using the Open MPI parallel environment ``mpi`` with a runtime of 30 mins and 2G of real memory per core.
+
+
+Installation notes
+------------------
+
+VASP 5.4.1 (05Feb16) was installed using the
+:download:`install_vasp.sh </sharc/software/install_scripts/apps/vasp/5.4.1.05Feb16/intel-17.0.0-openmpi-2.0.1/install_vasp.sh>` script;
+the module file is
+:download:`intel-17.0.0-openmpi-2.0.1 </sharc/software/modulefiles/apps/vasp/5.4.1.05Feb16/intel-17.0.0-openmpi-2.0.1>`.
```
dmrogers75: Aren't the module files named as per compiler or "binary"? So in the html doc it should read "VASP 5.4.1 (05Feb16)" and not "intel-17.0.0-openmpi-2.0.1"?

sharc/software/install_scripts/apps/vasp/5.4.1.05Feb16/intel-17.0.0-openmpi-2.0.1/install_vasp.sh
```diff
@@ -0,0 +1,82 @@
...
+#sudo='sudo'
+
+
+##############################################################################
+# This should not need modifying
+##############################################################################
+
+# Create the build dir
+
+if [ ! -d $build_dir ]
```
dmrogers75: Yes. There is no build dir specified. I will put one in the script, "/scratch/$USER/vasp", for future installs if the installer then only wishes to install the executables in $prefix.

sharc/software/apps/vasp.rst
anthonybrookfield: "mpirun vasp_std" should just work

sharc/software/apps/vasp.rst
anthonybrookfield: Yeah, the module filename would be intel-17.0.0-openmpi-2.0.1 but when it appears on the HTML doc the name doesn't make it clear that it's actually the VASP module.  

I'd probably do something like:
Install :download:`this modulefile </sharc/software/modulefiles/apps/vasp/5.4.1.05Feb16/intel-17.0.0-openmpi-2.0.1` as ``/usr/local/modulefiles/apps/vasp/5.4.1.05Feb16/intel-17.0.0-openmpi-2.0.1``

PR #681, opened 2017-07-18
Remove the "Source (GitHub)" link from the nav bar

Cadair:
The addition of this link has caused the search bar to re-flow off the nav for me:
screen shot 2017-07-18 at 13 07 42

and there is a "fork me on GitHub" link which points to the same place.


willfurnass:
@Cadair Ah, I see what you mean now!

Issue #493, opened 2016-12-13
ShARC: Install OpenMPI 1.10.4 w/ GCC 4.9.4 and ensure 1.10.4 properly configured


willfurnass:
See #483 for what is needed in `$MPI_HOME/etc/openmpi-mca-params.conf`

willfurnass:
The fix in #494 assumes that all out-of-band OpenMPI traffic can be routed over an interface called p3p1.321 and that an inteface of this name exists on all nodes that one might want to run MPI on.

However, on certain nodes (e.g. ShARC's public GPU-equipped nodes) the cluster-facing ethernet interface has a different name.  It may therefore make more sense to specify the interface to use for out-of-band comms using a (CIDR) network address e.g.

```
$ grep -E '^oob_tcp_if_include' /usr/local/packages/mpi/openmpi/1.10.4/gcc-4.9.4/etc/openmpi-mca-params.conf                                                 
oob_tcp_if_include = 172.30.64.0/22
```
This OpenMPI build passes the first stage of the Intel MPI Benchmarks; @gyengen is currently testing to see if this works for his workflow.

willfurnass:
@mikecroucher This fix has been implemented for only a subset of the OpenMPI 1.10.4 builds on ShARC:
```
/usr/local/packages/mpi/openmpi/1.10.4/gcc-4.9.4/etc/openmpi-mca-params.conf:oob_tcp_if_include = 172.30.64.0/22
/usr/local/packages/mpi/openmpi/1.10.4/gcc-6.2/etc/openmpi-mca-params.conf:oob_tcp_if_include = 172.30.64.0/22
/usr/local/packages/mpi/openmpi/1.10.4/intel-15.0.7/etc/openmpi-mca-params.conf:oob_tcp_if_include = p3p1.321
```
I'd be grateful if you could test the fix for the Intel build (if this is timely).  NB this is just a run-time fix; no recompilation is required.

If this fix works then the relevant `openmpi-mca-params.conf` in this repo and in production should both be updated.

mikecroucher:
Test directory copied at `/usr/local/packages/mpi/openmpi/1.10.4/intel-15.0.7_testing`. Created temporary module:

```
module load mpi/openmpi/1.10.4/testing123
```

Sent details of this to Ben Snow, an RA in maths who was trying his code out on the RSE queue. If it works for him, we are probably good to go.  Offered him a few weeks access to the RSE queue as a show of thanks for his help with testing.

[...]

PR #695, opened 2017-08-08
Fix Intel OpenMPI 1.10.4 out-of-band comms config

willfurnass:
Addresses #493


willfurnass:
@anthonybrookfield Fixed the broken cross-refs on this page.

anthonybrookfield:
cool - that's fixed it.

PR #716, opened 2017-09-28
gromacs on sharc: rst file, install script x2 and module file x2

dmrogers75:
GROMACS with and without GPU support installed on Sharc.


desmfryan:
Looks good. 

One point: from a performance point of view it would be good to understand the specification of GPU number and CPU number and whether there are default or recommended values i.e. if you specify nxGPU then you must specify mxCPU...what is the relationship between n & m (both integers obviously).

PR #763, opened 2018-01-05
Ensure newfloat.sty is built using sourcde file downloaded from CTAN, which should be more robust.

willfurnass:
FAO @twinkarma


willfurnass:
Need to merge in #764

PR #773, opened 2018-01-24
Updated job arrays page and tensorflow installation instruction.

twinkarma:
Does the section `Limiting number of concurrent tasks` sound ok? @willfurnass

### DIFF
--- a/parallel/JobArray.rst
+++ b/parallel/JobArray.rst
@@ -39,7 +39,7 @@ Array Jobs on ShARC and Iceberg can have a maximum of 75000 tasks.
 Limiting number of concurrent tasks
 -----------------------------------
 
-While unnecessary for CPU jobs, with the limited number of GPU nodes and how SGE has trouble accounting for fair use of GPUs, you may be asked to limit the number of jobs that will run concurrently.
+While unnecessary for CPU jobs, with the limited number of GPU nodes and how SGE has trouble accounting for fair use of GPUs, you may be asked to limit the number of GPU jobs that will run concurrently.
 
 Use the flag ``-tc 5`` to specify the maximum number of concurrent tasks (five tasks in this example). Same as the ``-t`` flag it can either be included on the command line: ::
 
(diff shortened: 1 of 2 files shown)


parallel/JobArray.rst
```diff
@@ -36,6 +36,23 @@ Advantages of Array Jobs:
 
 Array Jobs on ShARC and Iceberg can have a maximum of 75000 tasks.
 
+Limiting number of concurrent tasks
+-----------------------------------
+
+While unnecessary for CPU jobs, with the limited number of GPU nodes and how SGE has trouble accounting for fair use of GPUs, you may be asked to limit the number of GPU jobs that will run concurrently.
```
willfurnass: s/jobs/tasks/ ?

willfurnass: Suggested wording: "The job scheduler has means for ensuring fair use of the cluster if a user submits an array comprised of many CPU-only tasks.  However, if a user submits an array of GPU jobs then this could result in unfair usage as at present the job scheduler has mechanisms for lowering the priorities of pending tasks based on the user's recent CPU and memory usage but does _not_ penalise for recent GPU usage.  

Therefore, if you want to submit **GPU array jobs** you may **want (or be asked) to explicitly limit the number of GPU tasks that can run concurrently**. "

parallel/JobArray.rst
```diff
@@ -36,6 +36,23 @@ Advantages of Array Jobs:
 
 Array Jobs on ShARC and Iceberg can have a maximum of 75000 tasks.
 
+Limiting number of concurrent tasks
```
willfurnass: "...for GPU array jobs" ?

parallel/JobArray.rst
```diff
@@ -36,6 +36,23 @@ Advantages of Array Jobs:
...
+-----------------------------------
+
+While unnecessary for CPU jobs, with the limited number of GPU nodes and how SGE has trouble accounting for fair use of GPUs, you may be asked to limit the number of GPU jobs that will run concurrently.
+
+Use the flag ``-tc 5`` to specify the maximum number of concurrent tasks (five tasks in this example). Same as the ``-t`` flag it can either be included on the command line: ::
```
willfurnass: s/Same as/As per/

PR #779, opened 2018-02-02
Replace his/her with your

jezcope:
The rest of the document is already in second-person language (plus "his/her" seems awkward when the gender of the reader is irrelevant).


willfurnass:
Good spot.  Thanks @jezcope.

PR #781, opened 2018-02-07
Added FastQC and updated Foxit

gyengen:


desmfryan:
Hi Norbert

Despite the general straightforward nature of the FastQC install, it is still expected that we have a install script file even if it only contains 2 or 3 lines (and this should be linked in the documentation rst file).

The same goes for the module file...there should be a link to the module file, and the module file has the directory path name.

Foxit mods look fine (thanks).

Cheers,

Des

willfurnass:
Also, it's a good idea to create a pull request from a dedicated 'feature branch' rather than the `master` of your fork as otherwise you're blocked from making more commits your `master` until this pull request is approved and merged.

PR #782, opened 2018-02-08
bwa on sharc: rst file, install script and module file

dmrogers75:
For BWA recently installed on Sharc.

### DIFF
--- a/sharc/software/apps/bwa.rst
+++ b/sharc/software/apps/bwa.rst
@@ -1,0 +1,30 @@
+BWA
+===
+
+.. sidebar:: BWA
+   
+   :Version: 0.7.17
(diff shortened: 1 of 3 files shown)


sharc/software/apps/bwa.rst
```diff
@@ -0,0 +1,30 @@
+BWA
+===
+
+.. sidebar:: BWA
+   
+   :Version: 0.7.17
+   :Dependencies: Module loaded for GCC compiler 6.2.0
```
gyengen: Dependency:

dmrogers75: Dependeccies has been changed to Dependency.

PR #790, opened 2018-02-26
rosetta on sharc: rst file update

dmrogers75:


dmrogers75:
I assume Norberto is happy with this.

PR #792, opened 2018-03-02
remove references to iceberg gateway nat for remote-vis

anthonybrookfield:
No longer using private NAT, GPU-visualisation nodes are now accessed directly.  Remove references to iceberg-gateway.


willfurnass:
Covered by #788

PR #805, opened 2018-04-05
turbomole on sharc: rst file updated, 1 install script, 1 module file

dmrogers75:


dmrogers75:
Merged by DM Rogers as minor changes.

PR #759, opened 2017-12-20
CUDA + cuDNN changes for SHARC

willfurnass:
Install CUDA 9.1, CuDNN 7.0 for CUDA 9.1 and for CUDA 8.0:
- [x] write install scripts
- [x] write modulefiles
- [x] update docs
- [x] test - built and ran several CUDA example progs 
- [x] deploy

Also, 
- [x] note which versions of CUDA require which versions of GCC.
- [x] ensure that CUDA module files on ShARC are mutually exclusive.

PR #808, opened 2018-04-05
Revert "CUDA + cuDNN changes for SHARC"

willfurnass:
Reverts rcgsheffield/sheffield_hpc#759 - should wait on CUDA 9.1-compatible  NVIDIA driver to be installed on all ShARC GPU nodes before documentation on CUDA 9.1 goes live.


willfurnass:
Not needed; update of GPU drivers well under way.

PR #814, opened 2018-05-14
bump singularity version to 2.5.1 to match installed version on ShARC

anthonybrookfield:


anthonybrookfield:
self-merged 'cos just bumping the version number.

PR #821, opened 2018-06-06
Tidy ShARC docs on using R

willfurnass:
Changes:

 - nicer reStructuredText formatting (greater use of fixed width fonts)
 - minor rewording
 - switch to using semantic line endings


willfurnass:
Thanks @annakrystalli !

PR #839, opened 2018-10-18
 Added warning to TF docs on how it blocks all GPU memory on nodes

twinkarma:
... and how to resolve it with the deviceSelect.sh written by @ptheywood

### DIFF
--- a/sharc/software/apps/gpu/deviceSelect.sh
+++ b/sharc/software/apps/gpu/deviceSelect.sh
@@ -1,0 +1,19 @@
+#! /bin/bash
+
+# Function to set the value of CUDA_VISIBLE_DEVICES to the id of the least-utilised device in the system.
+# This will not set the device if CUDA_VISIBLE_DEVICES has already been set, to prevent issues on the ShARC DGX-1 (Node 126) or other private nodes which use the device allocation model
+function setCUDAVisibileDevices {
+# If there is no value of CUDA_VISIBLE_DEVICE
+if [ -z "$CUDA_VISIBLE_DEVICES" ]; then
+    # Query device info, to select the device id with the least chance of utilsiation. Sorts by free memory, then gpu utilsation, memory utilsiation, temperature and then id
+    deviceid=$(nvidia-smi --query-gpu=index,name,memory.free,utilization.gpu,utilization.memory,temperature.gpu --format=csv,noheader,nounits | sort -t, -k3,3nr -k4,4n -k5,5n -k6,6n -k1,1nr | head -n 1 | cut -d, -f1)
+    # Output and export the cuda visible devices variable
+    echo "Setting CUDA_VISIBLE_DEVICES=$deviceid"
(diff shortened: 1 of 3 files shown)


willfurnass:
Script doesn't account for the user requesting >1 GPU (or parallel jobs (MPI or SMP)). 

A cleaner approach is to enable the GPU locking epilog/prolog scripts on the public GPU nodes; this has been approved recently.  However, your doc suggestions would be useful in the interim if the aforementioned deficiencies in the script can be addressed (for the >1GPU and SMP job cases at least).

ptheywood:
I've created an updated version of the device selection script which can account for multiple devices including parallel jobs (thanks to @willfurnass). 

It also resolves a bug with the previous script, where the device indices returned by nvidia-smi are not the same as when enumerated as a part of `CUDA_VISIBLE_DEVICES` (via `CUDA_DEVICE_ORDER=PCI_BUS_ID`)

https://gist.github.com/ptheywood/ce4e5a600216dd085cae5bd5802d35aa

twinkarma:
Just ran the script and got an error

```
./selectCUDADevices.sh: line 13: 1 *  : syntax error: operand expected (error token is "*  ")

```

ptheywood:
I receive that error when I've not requested any GPUs, but with `#$ -l gpu=2` it runs as intended.

Additionally the call to nvidia-smi fails on non-gpu nodes, so we should probably account for that case and produce a graceful error, rather than 
```
/var/spool/sge/sharc-node025/job_scripts/2572589: line 46: nvidia-smi: command not found
```

ptheywood:
`getSGERequestedGPUCount` doesn't work in interactive sessions as `$JOB_ID` is unset.

ptheywood:
https://gist.github.com/ptheywood/ce4e5a600216dd085cae5bd5802d35aa#file-selectcudadevices-sh-L26 I've updated the gist to ensure that nvidia-smi is available so on non-gpu nodes (or if nvidia-smi is not available) an meaningful error message will be produced. This does not solve the interactive job issue.

ptheywood:
https://gist.github.com/ptheywood/ce4e5a600216dd085cae5bd5802d35aa

Now updated to correctly check for nvidia-smi, correctly handle gpu interactive sessions, and also handle qrsh based sessions (via @willfurnass).

twinkarma:
Thanks for sorting this out guys! Changes updated to the PR.

twinkarma:
@willfurnass are we happy with this? Can we start the merge?

twinkarma:
Looks like the script is working for both interactive and batch sessions now. Shall we go ahead with this on the interim before we can implement a better solution? @willfurnass

willfurnass:
As discussed, I'm keen to roll out the automatic setting of `CUDA_VISIBLE_DEVICES` ASAP as this will be much cleaner.

twinkarma:
Cancelling this as the automatic `CUDA_VISIBLE_DEVICES` makes this PR obsolete.

PR #857, opened 2018-11-27
Fix ansys modulefiles on Iceberg so do not load custom GL lib

willfurnass:
Attempts to run Ansys 16.1 and 17.2 on Iceberg using X forwarding or TigerVNC + VirtualGL fail with something like

```
ICEM_ACN is "/usr/local/packages6/ansys/ansys_inc/v161/icemcfd/linux64_amd".
args =
Window information: depth = 24
                    class = 4
visual depth = 24 class = 4
Xlib:  extension "NV-GLX" missing on display "iceberg-login3.shef.ac.uk:10.0".
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  149 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
  Value in failed request:  0x0
  Serial number of failed request:  4834
  Current serial number in output stream:  4835
Exit from ICEM CFD
```

*Not* prepending `LD_LIBRARY_PATH` with a custom GL lib addresses the issue.

### DIFF
--- a/iceberg/software/apps/ansys.rst
+++ b/iceberg/software/apps/ansys.rst
@@ -17,7 +17,8 @@ The Ansys suite of programs can be used to numerically simulate a large variety 
 
 Interactive usage
 -----------------
-After connecting to iceberg (see :ref:`ssh`),  start an interactive sesssion with the ``qsh`` command. Alternatively, if you require more memory, for example 16 GB, use the command ``qsh -l rmem=16G``.
+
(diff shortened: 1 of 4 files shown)


willfurnass:
Tested on Linux.  @me1bdh: please could you test on Windows.

me1bdh:
Tested on Windows. Appears to work.

PR #874, opened 2019-01-24
Update abaqus.rst

me1bdh:
Additional information added to prompt users to disable file locking env variable to prevent usage issues with /fastdata.


willfurnass:
@me1bdh Great tip!

PR #875, opened 2019-01-24
conda environment needs to be activated

drj11:
The documentation fails to point out that you need to activate the conda environment.


willfurnass:
Good spot, cheers.

PR #878, opened 2019-01-25
Document how to use Go

drj11:


willfurnass:
Mostly LGTM.  Some suggestions:
 - Make URLs hyperlinks? 
 - Any need to mention `GOROOT` / `GOPATH`? (the answer to this may be 'no')

drj11:
So, the whole `GOROOT` / `GOPATH` is something we avoid by using `conda`.
A later improvement to this document would to add the section: "ok, so _you_ know how Go is ordinarily organised; let's explain what conda has done".

willfurnass:
Fair enough.  Could you convert the URLs to hyperlinks?

drj11:
Yes, will add hyperlinks

drj11:
I converted one of the URLs to an ordinary hyperlink.

For the remainder I think it's useful to see and know the actual URL (and of course, in the rendered HTML document they appear as hyperlinks too).

Please reconsider merging.

PR #887, opened 2019-03-22
Add notes for Parallel overview

drj11:
Following conversation with @willfurnass in the RSE office, I made some notes.

### DIFF
--- a/parallel/index.rst
+++ b/parallel/index.rst
@@ -31,7 +31,7 @@ Computations running on cores on the same node can share memory.
 Code that runs on multiple cores may require that the cores are
 all on the same node or may not;
 additionally it may require that the code runs simultaneously on
-cores, or not.
+multiple cores, or not.
 This gives rise to a number of ways to use multiple cores:
 
 - all cores on same node: this is called `Shared Memory Parallelism <SMP.html>`_;
--- a/parallel/index.rst
+++ b/parallel/index.rst
@@ -8,12 +8,12 @@ A *core* (also called *processor*) is capable of executing one
 thread of computation.
 
 Modern computers contain more than one core;
-a typical laptop usually containing either 2 or 4.
-`Hyperthreading <https://en.wikipedia.org/wiki/Hyper-threading>`_
-is a way of excuting 2 (typically) threads on one
-core, it is enabled on most laptop-class cores,
-but is disabled on most HPC clusters
-(it is disabled on most nodes on ShARC).
+a typical laptop usually contains either 2 or 4.
+`Hyper-threading <https://en.wikipedia.org/wiki/Hyper-threading>`_
+is a way of excuting 2 (typically) threads on one core,
+it is enabled on most laptop-class cores,
+but is disabled on most HPC clusters.
+Hyper-threading is disabled on most nodes on ShARC.
 
 Computer clusters such as ShARC or Iceberg contain many hundreds of cores and
 the key to making your research code faster is to distribute your work across them.
--- a/parallel/index.rst
+++ b/parallel/index.rst
@@ -3,15 +3,47 @@
 Parallel Computing
 ==================
 
-Modern computers contain more than one processor with a typical laptop usually containing either 2 or 4 
-(if you think your laptop contains 8, it is because you have been fooled by `Hyperthreading <https://en.wikipedia.org/wiki/Hyper-threading>`_). 
-Computer clusters such as ShARC or Iceberg contain many hundreds of processors and 
-the key to making your research faster is to learn how to distribute your work across them. 
-If your program is designed to run on only one processor, 
-running it on our systems without modification will not make it any faster (it may even be slower!). 
+Parallel Computing uses more than one *core*.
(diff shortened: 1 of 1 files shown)


parallel/index.rst
```diff
@@ -3,16 +3,48 @@
 Parallel Computing
 ==================
 
-Modern computers contain more than one processor with a typical laptop usually containing either 2 or 4 
-(if you think your laptop contains 8, it is because you have been fooled by `Hyperthreading <https://en.wikipedia.org/wiki/Hyper-threading>`_). 
-Computer clusters such as ShARC or Iceberg contain many hundreds of processors and 
-the key to making your research faster is to learn how to distribute your work across them. 
-If your program is designed to run on only one processor, 
-running it on our systems without modification will not make it any faster (it may even be slower!). 
+Parallel Computing uses more than one core.
```
willfurnass: Too CPU-centric?  Also, can get parallelism at other levels e.g. via AVX SIMD.

parallel/index.rst
```diff
@@ -3,16 +3,48 @@
...
+A *core* (also called *processor*) is capable of executing one
+thread of computation.
+
+Modern computers contain more than one core;
+a typical laptop usually contains either 2 or 4.
+`Hyper-threading <https://en.wikipedia.org/wiki/Hyper-threading>`_
+is a way of excuting 2 (typically) threads on one core,
+it is enabled on most laptop-class cores,
+but is disabled on most HPC clusters.
+Hyper-threading is disabled on most nodes on ShARC.
```
willfurnass: Add a note that HT disadvantageous for numerical workloads?

parallel/index.rst
```diff
@@ -3,16 +3,48 @@
...
+Computer clusters such as ShARC or Iceberg contain many hundreds of cores and
+the key to making your research code faster is to distribute your work across them.
+If your program is designed to run on only one core,
+running it on an HPC cluster without modification will not make it any faster (it may even be slower!).
 Learning how to use parallelisation technologies is vital.
 
 This section explains how to use the most common parallelisation technologies on our systems.
 
+A CPU contains 1 or more cores.
+On ShARC the CPUs have 8 cores each.
```
willfurnass: Not always true.  Private notes are pretty heterogeneous.

willfurnass: True for CPUs in public nodes though.

parallel/index.rst
```diff
@@ -3,16 +3,48 @@
...
+A *node* on ShARC has 2 CPUs and so 16 cores.
+Computations running on cores on the same node can share memory.
+
+Code that runs on multiple cores may require that the cores are
```
willfurnass: "This is particularly true of programs with multiple threads that all need access to the same memory (memory cannot easily be shared between nodes)"

parallel/index.rst
```diff
@@ -3,16 +3,48 @@
...
+all on the same node or may not;
+additionally it may require that the code runs simultaneously on
+multiple cores, or not.
+This gives rise to a number of ways to use multiple cores:
+
+- all cores on same node: this is called `Shared Memory Parallelism <SMP.html>`_;
```
willfurnass: ", with one of the most common libraries for exploiting SMP being OpenMP"

parallel/index.rst
```diff
@@ -3,16 +3,48 @@
...
+- cores across many different nodes: `Message Passing Interface <MPI.html>`_;
```
willfurnass: MPI is the most common approach but not the only one (see also Spark, Hadoop, Dask etc).

parallel/index.rst
drj11: True, but with this pull request I wasn't intending to add any new discussion points, merely fill-in what was already there.

parallel/index.rst
drj11: yes, it would be reasonable to explain why Hyper-threading is disabled, or link to http://docs.iceberg.shef.ac.uk/en/latest/sharc/cluster_specs.html#general-cpu-node-specifications where it is also not explained. Extend the PR if you like.

parallel/index.rst
drj11: I'll fix that.

parallel/index.rst
drj11: extend the PR if you want to

parallel/index.rst
drj11: If you like. The SMP page says that, so I'm not sure if it needs repeating here.

parallel/index.rst
drj11: Right, but again I'm trying to fill in the existing material, not create new avenues for discussion.

parallel/index.rst
drj11: Created #888 instead.

parallel/index.rst
drj11: I've reworked this paragraph.

parallel/index.rst
drj11: Created #889 instead.

Issue #893, opened 2019-04-09
Install CUDNN 7.5 with CUDA 10.0 support


ptheywood:
The latest version of tensorflow requires CUDA 10.0 and an appropriate version of CUDNN (>= 7.4.1) built with CUDA 10.0

willfurnass:
Done; see PR #896

PR #896, opened 2019-04-11
ShARC: cuDNN 7.5.0.56 for CUDA 10.0 and 9.0

willfurnass:


willfurnass:
Addressed via PR #893

PR #901, opened 2019-04-11
Update binary

me1bdh:
Setting additional alias for APDL/Mechanical and ensure $ANSYS_DIR/ansys/bin is set on all modules as well as CFX root bin dir.


willfurnass:
@me1bdh Where is the Tcl variable `localroot` set?

me1bdh:
@willfurnass -  Ah. For versions > 17.2 `$localroot` should not be there. Shall amend.

PR #904, opened 2019-04-11
Update binary

me1bdh:
Setting additional alias for APDL/Mechanical and ensure $ANSYS_DIR/ansys/bin is set on all modules as well as CFX root bin dir.


me1bdh:
> Can't see where is $vers defined.

Defined line 12. Think this one was originally written by Deniz

PR #917, opened 2019-05-31
initial setup for bessemer docs and singularity install

anthonybrookfield:
Quick initial specs and structure for docs for bessemer (new HPC) so we can start populating docs.

(or would it be better to keep it all in a separate branch for now....?)


willfurnass:
+1 for separate `bessemer` branch in this repo: I've pushed your `bessemer` branch here.

PR #919, opened 2019-06-07
Warn that /tmp is automatically bind-mounted into Singularity containers on ShARC

willfurnass:
This *will* catch out folks who have installed software in `/tmp` when building Docker/Singularity images.  

Haven't yet found a workaround (the `-c` flag of `singularity exec` doesn't help).


willfurnass:
Tests failing due to broken Sphinx builds (addressed in PR #920)

PR #922, opened 2019-06-11
WIP: poetry support

willfurnass:
Potentially a simpler way to build and test locally than virtualenvs and requirements.txt

Need to check that works with Travis and RTD


willfurnass:
Experiment failed; will continue testing Poetry in another repo.

PR #925, opened 2019-06-19
Installed PGI 19.5, OpenMPI 4.0.1 (gcc and pgi), cctools 7.0.14 

gyengen:


willfurnass:
Need to issue pull request against `master`, not `sharc`, as the latter is no longer used.

PR #929, opened 2019-07-10
Bessemer pilot documentation

gyengen:
A general introduction of Slurm for pilot users.

### DIFF
--- a/bessemer/index.rst
+++ b/bessemer/index.rst
@@ -16,6 +16,8 @@ This is the documentation for Bessemer, The University of Sheffield's new High P
 
   software/index
   cluster_specs
+  slurm
 
 * :ref:`bessemer-software`
 * :ref:`bessemer-specs`
+* :ref:`slurm_info`
--- a/bessemer/slurm.rst
+++ b/bessemer/slurm.rst
@@ -1,0 +1,172 @@
+.. _slurm_info:
+
+Slurm Workload Manager
+======================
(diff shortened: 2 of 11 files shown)


anthonybrookfield:
I'm still in two minds as to whether it's better to have a dedicated Slurm page for bessemer, or if it's better to include Slurm commands in the top-level 'Using interactive sessions/submitting batch jobs' HPC section.  I've tried including slurm-equivalent commands in the hpc/submit.rst file to replace the hpc/sge.rst file - on the plus side it's easier for users to see equivalent commands if they're familiar with SGE, on the other hand it's more complicated for new users.  

However if we did have a separate Slurm page we'd need a separate SGE page and then the top level page rewriting/removing...... The top level page does include useful generic info on interactive/batch jobs which we don't really want to repeat on SGE and SLURM specific pages.

I'm willing to be convinced either way.

General comment - I think the examples need to differentiate between parallel job types, e.g. MPI-style jobs which would use the ntasks option and SMP style -c / --cpus-per-task option.  Also at least initially we'll be enforcing MPI jobs to sit within a single node, so max job size will be 40 cores.

gyengen:
Since Bessemer is going to be our general cluster (= more inexperienced users) I would not mix the SGE and Slurm commands.

gyengen:
Pilot documentation for the installed software is uploaded.

PR #931, opened 2019-07-16
Add JADE to the cluster listing

twinkarma:
With information on how to request access


sharc/software/apps/gpu/selectCUDADevices.sh
```diff
@@ -0,0 +1,94 @@
+#! /bin/bash
```
willfurnass: Doesn't appear to be referenced anywhere in the docs.

sharc/software/apps/gpu/deviceSelect.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/bash
```
willfurnass: Doesn't appear to be referenced anywhere in the docs.

PR #949, opened 2019-09-06
Fix links to RSE blog posts

ptheywood:
A maple user pointed out the link to the tutorial on the RSE blog was incorrect. 

I have fixed the three incorrect links.

This is likely due to the RSE blog migrating form nikola to jekyll (which is case sensitive or only uses lower case slugs)


anthonybrookfield:
looks fine to me.  I'll merge.

Issue #973, opened 2019-11-20
Bessemer: Install CUDA 10.0


ptheywood:
PGI 19.1 (installed on bessemer as PGI/19.1-GCC-8.2.0-2.31.1) does not link against CUDA 10.1 codes successfully, as it only supports up to CUDA 10.0. The version of the cuda toolkit shipped with the PGI compiler is incomplete, so cannot be used explicitly.

Tensorflow ` >= 1.13.0` will also require CUDA 10.0 rather than 10.1, which will likely be in high demand once additional GPU nodes are installed.

willfurnass:
`CUDA/10.0.130` and `cuDNN/7.4.2.24-CUDA-10.0.130` modulefiles now available on Bessemer.  @ptheywood Think it's worth adding a note to https://docs.hpc.shef.ac.uk/en/latest/bessemer/software/development/pgi.html about CUDA compat?

ptheywood:
It might be worth it, though I think I ran into another issue with having to downgrade the GCC module for CUDA 10.0 nvcc to be happy to get my code to build. 

The `PGI` module adds `GCCCore/8.2.0`, but `CUDA/10.0.130` is incompatible with GCC 8 (4.8.5 is the officially supported version on CentOS 7, but 7.x appears to work. (although my application doesn't work, but thats more likely due to interesting fortran and changes in PGI).

willfurnass:
Think that can be closed now.

PR #974, opened 2019-11-20
Bessemer: basic user docs for CUDA 10.0 and cuDNN 7.4 for CUDA 10.1 and 10.0

willfurnass:
Should address #973 once installs have been tested


willfurnass:
Now tested

PR #1002, opened 2020-05-05
ShARC: CUDA/cuDNN updates and modulefile fixes

willfurnass:
- Installer/modulefile/docs for CUDA 10.1 and 10.2
- Installer/modulefile/docs for compatible version of cuDNN
- Tweak all CUDA modulefiles to 
    - set `CUDA_ROOT`, `CUDA_PATH` and `CUDA_HOME` to the CUDA install prefix (previously only `CUDA_HOME` set)
    - prepend paths to `CPATH` and `MANPATH`
- Note that CUDA 10.1 required by TensorFlow 2.1

Addresses #884, #897 and partly addresses #999

PR #1004, opened 2020-05-07
Changed default installed tf gpu version to 2.0

twinkarma:
Added tf versions to compatibility matrix


sharc/software/apps/tensorflow.rst
```diff
@@ -120,11 +120,17 @@ Tensorflow releases depend on specific versions of both CUDA and CUDNN. If the w
...
 
-The following table shows the which module to load for the various versions of Tensorflow, based on the `tested build configurations <https://www.tensorflow.org/install/source#linux>`_. Newer versions may require more recent CUDA and CUDNN releases. 
+The following table shows the which module to load for the various versions of Tensorflow, based on the `tested build configurations <https://www.tensorflow.org/install/source#linux>`_. 
+
+
 
 +------------+------+--------+--------------------------------------------+
 | Tensorflow | CUDA | CUDNN  | Module                                     | 
 +============+======+========+============================================+
+| >= 2.0.0  | 10.0  | >= 7.4 | `libs/cudnn/7.5.0.56/binary-cuda-10.0.130` |
```
willfurnass, line 130: Does tensorflow 2.1.0 require CUDA 10.1?

twinkarma: Yes it does

twinkarma: I've removed the `>=` for the latest three versions but I don't really want to add 2.1.0 yet as we don't have a CUDA 10.1 module

ptheywood: Would it not be useful to say that is the case but this is  not yet available?

willfurnass: Added TF 2.1 requirements in PR #1002

PR #1006, opened 2020-05-11
Docs for using Spark with JupyterHub on ShARC

willfurnass:
Here I suggest loading Java via an environment module but installing Java via conda may be slicker (but a bit more bloaty) - Java via conda needs testing.


willfurnass:
Tested installing `openjdk` using conda but pyspark couldn't find it, most likely because of nb_conda_kernels-related redirection/magick.  Added a note to the docs that using openjdk with conda on ShARC may work with some effort but has not been tested.

PR #1021, opened 2020-05-28
Use Sphinx RTD theme?

willfurnass:
I think this looks better than our customised `sphinx-bootstrap-theme`.  @ptheywood and @twinkarma - any opinions?


ptheywood:
I would say I slightly prefer the RTD theme than the modified bootstrap theme.
Espeically at 1124px wide (1/3 of my 3440px monitor) where the search box esacpes the nav bar in the sphinx-bootstrap version. 

The group GPU node pages are looking good, though you've got the COM nodes down as 32GB cards? (as far as I'm aware they were going to be 16GB cards like in the academics nodes).

 > 4x NVIDIA Tesla V100 SXM2 (32GB RAM each; NVLINK interconnects between GPUs)

willfurnass:
@ptheywood They're definitely 32GB devices :)

twinkarma:
I think the RTD theme looks good but I think that the side navigation bar's font on the 4th level is too small? e.g. if you look at:

ShARC -> Using GPus on Sharc -> ShARC GPU Resources -> Hardware

willfurnass:
@twinkarma Possibly.  That leads neatly on to another question: how much of the stuff in `themes/tuos` do we need to retain if switching to using `sphinx-rtd-theme`?  We can specify a logo directly in `conf.py` and that logo presumably doesn't need to live beneath `themes`.

twinkarma:
Not sure what you mean, I see that you've switched the base theme in themes/tuos to sphinx-rtd them, is it not ok to just leave it as it is? Did you want to remove the tuos theme folder completely?

jkwmoore:
I do like it quite a bit - though we might want to align centre?  

image

willfurnass:
Rebased on `master` and renamed our custom theme (mixture of `iceberg` and `tuos` -> `sheffieldhpc`).

@twinkarma I've made the l4 headings bigger as requested.

@jkwmoore Thanks for the suggestion; the layout is now wider.

jkwmoore:
Looking good to me!

Issue #1034, opened 2020-06-22
Add placeholder info re JADE II


willfurnass:
@twinkarma Would you be happy drafting something that includes the following?
 - relevant research domains
 - key architectural features
 - spec (hardware, OS, software stack (if known yet))
 - whether timeline for researcher access known yet.

PR #1036, opened 2020-06-22
Added a page on JADE 2 as requested in #1034

twinkarma:


jade2.rst
```diff
@@ -0,0 +1,36 @@
...
+
+**More details on the system coming soon.**
+
+Hardware
+^^^^^^^^
+
+* 63 Nodes of DGX MAX-Q, each with:
+    * 8x Nvidia V100 32GB GPUs
+    * 512GB RAM
+    * 70TB
```
willfurnass, line 26: 70TB local storage per node?

twinkarma: Oops I did a copy instead of cut

jade2.rst
```diff
@@ -0,0 +1,36 @@
...
+**More details on the system coming soon.**
+
+Hardware
+^^^^^^^^
+
+* 63 Nodes of DGX MAX-Q, each with:
+    * 8x Nvidia V100 32GB GPUs
+    * 512GB RAM
+    * 70TB 
+* 70TB DDN AI400 NVMe storage
```
willfurnass, line 26: This suggests it's NVMe storage array(s) that run a parallel filesystem and are accessed over the EDR Infiniband cluster fabric - is this right?

jade2.rst
```diff
@@ -0,0 +1,36 @@
...
+* 1PB Lustre disk storage
+* EDR infiniband interconnect
```
willfurnass, line 28: For connections to storage or for inter-node comms too?

jade2.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Hardware
+^^^^^^^^
+
+* 63 Nodes of DGX MAX-Q, each with:
+    * 8x Nvidia V100 32GB GPUs
+    * 512GB RAM
+    * 70TB 
+* 70TB DDN AI400 NVMe storage
+* 1PB Lustre disk storage
```
willfurnass, line 27: Worth noting that Lustre is a shared filesystem that (presumably) can be accessed by all nodes in the cluster.

jade2.rst
twinkarma: It should be infiniband for everything.

willfurnass: So multi-node jobs will be permitted with JADE II?

jade2.rst
twinkarma: Yes the 70TB NVMe and 1PB storage should both be network storage. One for fast read/write and one for normal storage/archive.

willfurnass: Interesting.  I think that comment would be worth adding to this page.

jade2.rst
twinkarma: I've heard it mentioned that its on the cards but will confirm again.

willfurnass: I guess it's probably wise not to mention the possibility of  multi-node jobs here until that's been confirmed.

willfurnass:
@twinkarma OK to merge?

twinkarma:
Yes let's merge it in then, I won't be getting any new info in the nearish future.

willfurnass:
Thanks @twinkarma !

PR #1047, opened 2020-07-10
Bessemer: remove references to non-existant cmake builds

### LINKED ISSUE #1046: CMake docs for Bessemer reference non-existant modulefiles
Compare https://docs.hpc.shef.ac.uk/en/latest/bessemer/software/development/cmake.html and

```
$ ma | grep CMake
CMake/3.13.3-GCCcore-8.2.0
```
All but one of the versions of CMake listed on that page are only installed in the software staging area but not the live area.

willfurnass:
Fixes #1046


willfurnass:
That's more like it:
```
$ ma | grep -i cmake
CMake/3.12.1-GCCcore-7.3.0
CMake/3.13.3-GCCcore-8.2.0
CMake/3.15.3-GCCcore-8.3.0
```

cc @ptheywood

PR #1063, opened 2020-07-29
clarify GPU memory request

hertzsprung:
where K80 has 12GB onboard memory, qrshx command is requesting 13GB of main memory


willfurnass:
Thanks @hertzsprung

PR #1064, opened 2020-07-31
Add R on Bessemer

willfurnass:
It's just R built using / depending on the `foss-2019b` toolchain.  No 3rd-party R pkgs installed.  Can install R packages such as rgdal and rgeos etc in due course.  Can also install R built with an `intel` toolchain if requested too.

### DIFF
--- a/bessemer/software/apps/R.rst
+++ b/bessemer/software/apps/R.rst
@@ -1,0 +1,177 @@
+.. _bessemer_r:
+
+R
+=
+
+.. sidebar:: R
+
+   :URL: https://www.r-project.org/
+   :Documentation: https://www.r-project.org/
+
+R is a statistical computing language.
+
+Interactive Usage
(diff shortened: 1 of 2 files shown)


willfurnass:
Still WIP as a dependency (a Python easyconfig) failed to build...

jkwmoore (commented):
All seems fine, only comment is whether we want to match the libraries on ShARC and make further builds compiled with the Intel Compiler available (I know this is likely non-trivial.)

willfurnass:
> All seems fine, 

Ta.

> only comment is whether we want to match the libraries on ShARC and make further builds compiled with the Intel Compiler available (I know this is likely non-trivial.)

For libs such as rgdal etc?

willfurnass:
Can add in due course.  Think it's important to get this merged now.

PR #1067, opened 2020-08-11
add instructions for intel compilation under cmake

hertzsprung:
I hope I got the hyperlink syntax correct, I'm not too familiar with rst/sphinx notation!


willfurnass:
Thanks @hertzsprung!

PR #1070, opened 2020-08-25
ShARC/Iceberg: note restricted perms on central conda install

willfurnass:
Not needed on Bessemer as EasyBuild configured to remove write perms after installing each easyconfig in the live area.


jkwmoore (commented):
All good, but I would probably clarify this also prevents the weird deadlock on env creation in their home dirs also!

Issue #1082, opened 2020-09-16
Update Bessemer Fluent instructions to include disabled affinity.


jkwmoore:
Investigation is ongoing however - it appears that allowing Fluent to set its own core affinity leads to it failing to allocate program threads to cores sensibly and duplicating multiple program threads on one core. 

This results in performance losses as some cores will be left idle out of the reservation.

ANSYS inform us the fix is as follows:

 

**SLURM is not officially supported but we have seem similar affinity issues before with running Fluent. The best option would be to run Fluent with the -affinity=off flag set. This prevents Fluent from setting affinities and leaves your clusters operating system to allocate cores to nodes. If you are not already doing this, please try this first.

I only know about affinity setting in CFX, Fluent and Mechanical - APDL. The software tries to set sensible defaults that give the most efficient performance but, very occasionally, that can clash with the affinity setting preferred by the cluster manager.**

This fix has been tested and appears to work correctly on Bessemer so documentation needs to be updated, see below:

![image](https://user-images.githubusercontent.com/69149348/93330036-0e3bf200-f816-11ea-9723-cd0c2d8e1b03.png)
![image](https://user-images.githubusercontent.com/69149348/93330055-15630000-f816-11ea-94f8-3a0c7823b0bf.png)

jkwmoore:
Considering the following tests from ShARC - SGE does not appear to have this issue in either MPI or SHM modes:

![image](https://user-images.githubusercontent.com/69149348/93351526-c1661480-f831-11ea-800c-1645dcb25e84.png)

![image](https://user-images.githubusercontent.com/69149348/93351533-c4f99b80-f831-11ea-9da6-30e839292019.png)

jkwmoore:
The affinity fix has been tested for MAPDL also to verify jobs will run, but no benchmarking has been undertaken.

jkwmoore:
The fix for this issue is as follows - export FLUENT_AFFINITY=0 is being added to the module files to ensure that Fluent allocates processes to cores correctly.

Testing of all versions of Fluent on Bessemer has been completed with the above successfully and fixes the issue.

No fix is required for MAPDL as confirmed with ANSYS, KMP_Affinity is being dealt with via another issue and PR!

jkwmoore:
Live testing of the corrected module files is now being undertaken.

jkwmoore:
Further PR to add module files to documentation here: https://github.com/rcgsheffield/sheffield_hpc/pull/1099

[...]

PR #1085, opened 2020-09-17
Update to reflect optimal affinity settings for Bessmer

jkwmoore:
Update adds the -affinity=off to the commands where needed and provides explanation for this setting.

This is the PR for https://github.com/rcgsheffield/sheffield_hpc/issues/1082


willfurnass:
Can we set this as a default for those applications somehow?

jkwmoore:
Waiting on a reply from ANSYS :)

jkwmoore:
As per ANSYS' reply - fix is proposed instead to use the tested env var rather than asking used to use command flags.

Amended as such - changes to module files are pending (likely tomorrow morning) subject to change request being approved / implemented.

desmfryan (approved):
Hi James

Looks fine. I will run a quick Fluent job on Bessemer as a pretend user :+1: 

Cheers,

Des

PR #1086, opened 2020-09-25
Update docs to reflect the KMP_Affinity settings

jkwmoore:
Update to clarify/detail why the KMP_Affinity setting is needed only for ANSYS 19.X for both Bessemer and ShARC.

The files on the cluster have been edited to reflect this with the settings removed for ANSYS 20.X

Minor style edits also.

### DIFF
--- a/bessemer/software/apps/ansys.rst
+++ b/bessemer/software/apps/ansys.rst
@@ -101,7 +101,7 @@ The job is submitted to the queue by typing::
 Installation note for Administrators:
 -------------------------------------
 
-MAPDL will not run in ANSYS versions 19.X without modifying the file::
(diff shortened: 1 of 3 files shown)


jkwmoore:
Closing this PR as despite an ANSYS rep telling us this was fixed it has not been and still requires the following fix inserting:

```
The following instruction should be inserted at line 2433 in anssh.ini:

setenv KMP_AFFINITY compact

```
This was missed during initial testing after this patch was removed (~10 batch jobs) as their core allocation was luckily correct and no jobs failed.

A while later user reported having issues which lead to investigation showing that MAPDL batch jobs would fail when their core allocation was wrong and interactive sessions with MAPDL would fail outright when requested.

The KMP patch has been reinstated on ANSYS 20.X on both clusters and current docs are correct due to this, hence this PR no longer needs to exist.

Issue #1079, opened 2020-09-10
SLURM memory request info is wrong on Submit Queue page


jkwmoore:
https://docs.hpc.shef.ac.uk/en/latest/hpc/scheduler/submit.html#submit-queue

![image](https://user-images.githubusercontent.com/69149348/92725796-90cf3800-f364-11ea-8b69-53430a08b5be.png)

The SLURM scheduler should be handling memory requests as a total, not per core! This will need a fix and other iterations of this also elsewhere (if present) will need a fix.

PR #1093, opened 2020-10-09
Clarify memory options in tables.

jkwmoore:
See: https://github.com/rcgsheffield/sheffield_hpc/issues/1079


willfurnass (commented):
I agree that this section would benefit from revising but think it might help to introduce the `--mem-per-cpu` argument as this works in the same way as SGE's `rmem'.  Could also mention `--mem-per-gpu`.

jkwmoore:
We might want to refactor the table on this page leaving blank section for comparison options where there is no clear alternative between SGE and SLURM - e.g. by SLURM by CPUs (not by node or by core) has no direct equiv in SGE (that I know of.)

jkwmoore:
I'm going to close this as part of a larger refactoring is planned.

jkwmoore:
I am going to re-open and merge this because that refactoring hasn't happened yet and I really don't want the docs staying wrong for any more time.

jkwmoore:
Currently looks like - 
image

desmfryan (commented):
Hi James

Some comments/suggestions:
-> means 'change to'

line 57 
to stat -> to start

line 129
-pe <env> -> -pe mpi

line 165
will be consumed -> will be run/executed 
Note: consumed will not mean anything to the general user

line 178:
consumed -> run/executed

line 261:
foo < foo.dat > foo.res      ->       foo foo.dat foo.res
note users may think they need to use < > around input files

line 268:
foo < foo.dat > foo.res      ->         foo foo.dat foo.res

line 363:
--mail-type=          ->      --mail-type=all

Cheers,

Des

jkwmoore:
> 
> 
> Hi James
> 
> Some comments/suggestions:
> -> means 'change to'
> 
> line 57
> to stat -> to start
> 

Fixed.

> line 129
> -pe -> -pe mpi
> 

I suspect this has been set as  -pe <env> because we have mpi and mpi-rsh envs. (ANSYS has to be special ofc.)

> line 165
> will be consumed -> will be run/executed
> Note: consumed will not mean anything to the general user
> 

Fixed.

> line 178:
> consumed -> run/executed
> 

Fixed.

> line 261:
> foo < foo.dat > foo.res -> foo foo.dat foo.res
> note users may think they need to use < > around input files
> 

Fixed.

> line 268:
> foo < foo.dat > foo.res -> foo foo.dat foo.res
> 

Fixed. (But was 286 iirc.)


> line 363:
> --mail-type= -> --mail-type=all
> 

Fixed.

PR #1101, opened 2020-10-16
Warning note added for Iceberg Decommissioning

jkwmoore:
Warning note added to all Iceberg pages - do we wish to add this to other pages?

### DIFF
--- a/iceberg/GPUComputingIceberg.rst
+++ b/iceberg/GPUComputingIceberg.rst
@@ -1,3 +1,10 @@
+.. Warning:: 
+    Iceberg reaches end-of-life on **30th November 2020.**
+    If you are running jobs on Iceberg then you need to take urgent action to ensure that your jobs/scripts will run on ShARC or Bessemer. 
+ 
+    If you have never used ShARC or Bessemer then now is the time to test your scripts.
+    Not all software on Iceberg is available on ShARC/Bessemer. 
+
 .. _GPUComputing_iceberg:
 
 Using GPUs on Iceberg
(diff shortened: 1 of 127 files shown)


desmfryan:
Hi James

Looks good to me.

Cheers,

Des

willfurnass:
Nice idea.  Cleaner to add the warning via a separate file referenced by the .reStructuredText 'include' directive?  Would more easily allow that warning to be updated.  See https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html#include-a-shared-file

willfurnass (commented):
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html#include-a-shared-file

jkwmoore:
> Caution
>Include paths are relative to the file in the document project, not the file in shared content.


Unfortunately including files is relative to the file you are compiling at the time so depending on where that .rst file is in the folder hierarchy will need a different file path for the include which is a non-trivial thing to fix.

I've tested this and it is indeed relative - we could ofc write something to parse what level of directory it is in and correctly add the  include path but it's Friday at 17:13 so I'm not going to look at that now and it probably isn't worth the time anyway :)

willfurnass:
Ah.  You could have one warning.rst in `iceberg/` then a different relative `.. include::` path in each of the following?
 
1. `iceberg/software/*/*.rst`
1. `iceberg/*/*.rst`
1. `iceberg/*.rst`

jkwmoore:
That should be using include files everywhere that is needed but there's going to be a lot of warnings when it generated the files.

jkwmoore:
Did I just kill the builder with excessive warning messages? :/ Built it fine locally.

PR #1112, opened 2020-11-03
Minor correction to V100 memory on interactive GPU examples

ptheywood:
This applies to the V100 node(s) in the gpu and dcs-gpu partitions, but not the 16GB V100s in the dcs-acad partition, which is an special case with limited access.

### DIFF
--- a/bessemer/GPUComputingBessemer.rst
+++ b/bessemer/GPUComputingBessemer.rst
@@ -35,10 +35,10 @@ As such, it is recommended that you request enough CPU memory to communicate pro
 
 .. code-block:: sh
 
-   # NB Each NVIDIA V100 GPU has 16GB of RAM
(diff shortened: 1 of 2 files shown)


willfurnass (changes requested):
Hi @ptheywood. Good spot.  See small suggestion in-line.  

Also, I think we should add a brief note to `bessemer/groupnodes/dcs-acad-gpu-nodes.rst` making folks aware that most other GPU nodes in the cluster have 32GB/device.

willfurnass:
@ptheywood I've suggested some changes for addressing these comments: https://github.com/ptheywood/sheffield_hpc/pull/1

ptheywood:
@willfurnass I've merged your suggested changes into the branch/pr.

willfurnass (approved):

willfurnass:
Thanks @ptheywood

PR #1126, opened 2020-12-16
Update Bede page now N8 CIR docs and Bede both live

willfurnass:
- Bede now live so need to use present tense
- Refactored to state most suitable workflows at top of page
- Simplified specifications
- Signposted to new Bede docs

### DIFF
--- a/bede.rst
+++ b/bede.rst
@@ -26,11 +26,13 @@ Noteworthy features of the system
   * 2x IBM POWER9_ CPUs 
   * 2x `NVIDIA V100`_ GPUs per CPU
   * Each CPU is connected to its two GPUs via high-bandwidth, low-latency interconnects (NVLink), which helps if you need to move lots of data to/from GPU memory
+  * 512 GB RAM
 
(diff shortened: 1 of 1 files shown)


willfurnass:
@ptheywood @mikeg64 Going to merge this before the RITF today.  If you got chance to review before then that would be great but no worries if not.

ptheywood (approved):
Looks good to me. 

Is it worth mentioning host memory (just to match the bede docs, 512GB for v100 nodes, 256GB for infer nodes)?

Some slightly odd margins for the nested lists, though that's a general comment (and I think v. common for rtd)

willfurnass:
@ptheywood Thanks for the suggestion re RAM/node; updated the PR.  Not going to faff with .rst formatting / CSS with this PR given it's important to get it merged before the RITF.

mikeg64:
Take care Bede is live but at Sheffield it's in beta phase - still getting processes in place - testing - allowing some beta users

willfurnass:
@mikeg64 Added caveats to that effect via PR #1127.

PR #1130, opened 2021-01-06
ShARC R GCC (not Intel) documentation update for R 4.0.0.

jkwmoore:
Install scripts, module files added.

Slight modification to the docs to remove latest version can be loaded with section to avoid people using incorrect versions.

.version file added to the R module top directory to ping this at version 3.6.4 to keep using this version while version 4.0.0 modules are tested by end users.

Important to note that R 4.0.0 now requires the use of PCRE2 which was also compiled and added to the cluster.

### DIFF
--- a/sharc/software/install_scripts/apps/R/4.0.0/gcc-10.1.0/install-R4.0-gcc-10.1.0.sh
+++ b/sharc/software/install_scripts/apps/R/4.0.0/gcc-10.1.0/install-R4.0-gcc-10.1.0.sh
@@ -1,0 +1,44 @@
+##!/bin/bash -e
+##$ -l rmem=40G
+
+#Set up environment variables and create directories
+version=4.0.0
+install_dir=/usr/local/packages/apps/R/$version/gcc-10.1/
+build_dir=~/R-$version-build
+mkdir -p $build_dir
+mkdir -p $install_dir
+module load dev/gcc/10.1
+cd $build_dir
+
+#Download the installer
+wget https://cran.r-project.org/src/base/R-4/R-$version.tar.gz
+tar -xzf ./R-$version.tar.gz
+cd R-$version
+
+tcl_config=/usr/local/packages/apps/tcl/8.6.10/gcc-10.1/lib/tclConfig.sh
+tk_config=/usr/local/packages/apps/tk/8.6.10/gcc-10.1/lib/tkConfig.sh
+
+#Configure and build
+./configure --prefix $install_dir --with-tcltk --with-tcl-config=$tcl_config \
+ --with-tk-config=$tk_config --with-blas --with-lapack \
+--enable-R-shlib 2>&1 | tee config-R-$version.log
+make 2>&1 | tee make-R-$version.log
+
+#Check
+make check 2>&1 | tee make_check-R-$version.log
+
+#Install test directory
+make install-tests
+
+#Install
+make install
+
+#Build libRmath.so
+cd $build_dir/R-$version/src/nmath/standalone
+make 
+mv libRmath.* $install_dir/lib64/R/lib
+
+#Copy install,configure and test log files
+mkdir -p $install_dir/install_logs
+cd $build_dir/R-$version
+mv ./*.log $install_dir/install_logs
--- a/sharc/software/install_scripts/apps/R/4.0.0/gcc-10.1.0/install-R4.0-gcc-10.1.0.sh
+++ b/sharc/software/install_scripts/apps/R/4.0.0/gcc-10.1.0/install-R4.0-gcc-10.1.0.sh
@@ -8,6 +8,7 @@ build_dir=~/R-$version-build
 mkdir -p $build_dir
 mkdir -p $install_dir
 module load dev/gcc/10.1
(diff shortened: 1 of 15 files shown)


willfurnass:
Still marked as draft - ready for review?

jkwmoore:
> 
> 
> Still marked as draft - ready for review?

Yep, just flipping that off draft now :)

jkwmoore:
*Note - we need to rename the module files when this is PR is merged!

sharc/software/modulefiles/apps/R/4.0.0/gcc-10.1.0
```diff
@@ -0,0 +1,21 @@
...
+
+set version 4.0.0
+module load dev/gcc/10.1
+module load libs/pcre2/10.36/gcc-10.1
+set R_DIR /usr/local/packages/apps/R/$version/gcc-10.1
+
+prepend-path PATH $R_DIR/bin
+prepend-path LD_LIBRARY_PATH $R_DIR/lib64/R/lib/
+
+setenv TK_LIBRARY /usr/local/packages/apps/tk/8.6.10/gcc-10.1/lib/tk8.6
```
willfurnass, line 20: Are the Tcl + Tk dependencies managed via compile-time args and these envs vars?  Not via modulefiles?  Might be worth adding modulefiles to make the libs easier to reuse for other builds.

sharc/software/install_scripts/apps/R/4.0.0/gcc-10.1.0/install-R4.0-gcc-10.1.0.sh
```diff
@@ -0,0 +1,45 @@
...
+#Check
+make check 2>&1 | tee make_check-R-$version.log
+
+#Install test directory
+make install-tests
+
+#Install
+make install
+
+#Build libRmath.so
```
willfurnass, line 37: Is libRmath an alternative to using an external lib for certain math operations?  Would an external lib give better perf?

willfurnass (commented):
See in-line comments.

Planning Intel counterpart(s) to these builds?

sharc/software/modulefiles/apps/R/4.0.0/gcc-8.2.0
```diff
@@ -0,0 +1,23 @@
...
+
+proc ModulesHelp { } {
+      global helpmsg
+      puts stderr "\t$helpmsg\n"
+}
+
+set version 4.0.0
+
+module load dev/gcc/8.2
+module load libs/pcre2/10.36/gcc-8.2.0
```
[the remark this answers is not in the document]
jkwmoore: Yes,  iirc R 4.0.0 requires it.

sharc/software/modulefiles/apps/R/4.0.0/gcc-10.1.0
jkwmoore: I had considered this, we can certainly do that but in my discussion with Des he'd said that with dependents like this typically were not module'd.

I'm not sure on the compile time args vs envs as I was following the previous module file's design.

sharc/software/install_scripts/apps/R/4.0.0/gcc-10.1.0/install-R4.0-gcc-10.1.0.sh
jkwmoore: A good question - we should probably ask @annakrystalli

willfurnass: Could also look at the mandatory/optional logic for this in the Spack and EasyBuild recipes

sharc/software/modulefiles/apps/R/4.0.0/gcc-10.1.0
willfurnass: Tcl/Tk are likely to be dependencies of other things so I think in these cases it makes sense to expose via modulefiles.

sharc/software/modulefiles/apps/R/4.0.0/gcc-8.2.0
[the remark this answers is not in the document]
willfurnass: Could you add to this PR then?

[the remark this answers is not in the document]
jkwmoore: Adding in latest commit!

sharc/software/modulefiles/apps/R/4.0.0/gcc-10.1.0
jkwmoore: Modules now added to commit - uploading to the cluster now and testing.

jkwmoore: Everything seems in working order.

jkwmoore:
@desmfryan Do you want to add the updates for R version 4.0.2 and 4.0.3 to this PR or shall I just merge?

PR #1132, opened 2021-01-13
Removing invalid line -  At present public GPUs are only available in batch jobs.

jkwmoore:
Maybe we edit the docs, maybe we edit the job_submit.lua - who knows!


willfurnass (approved):
@jkwmoore Let's merge this so that the docs accurately reflect the current Slurm config then subsequently discuss whether we need to prohibit interactive use of public GPUs.

PR #1133, opened 2021-01-18
ANSYS 2021 R1 update

jkwmoore:
Updates both ShARC and Bessemer docs.

Currently testing installs with Fluent and MAPDL.

Will change from draft as soon as testing is complete.


jkwmoore:
This cannot be merged until the 2021 licence server has been installed due to changes in licencing for the 2021 and subsequent releases.

jkwmoore:
This is now pending validation / verification that the module is working correctly.

Issue #1124, opened 2020-12-10
Add Miniconda installation documentation to the Python pages


jkwmoore:
This needs adding (mostly cut and paste from a ticket) but should ideally be put onto the existing Python Anaconda pages as a header section separated with a link from each cluster's top level software page.

jkwmoore:
Example spiel below:

You can see the available installers here: https://docs.conda.io/en/latest/miniconda.html

I will be using the Python 3.8 installer as the example. You should probably install this to your data directory i.e. /data/$USER/miniconda or a similar subdirectory as your home directory has limited space (ShARC-10GB and Bessemer-100GB) and conda packages can pile up quickly. Running out of space is sub-optimal!

#Download the installer
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh

#Check the sums match with those listed on the website
sha256sum Miniconda3-latest-Linux-x86_64.sh

#Make the file executable
chmod +x Miniconda3-latest-Linux-x86_64.sh

#Run the installer - Make sure to choose your data directory  and subdirectory i.e. /data/$USER/miniconda (don't make this directory prior to the installer as it makes the directory itself when you supply it with a directory path.)
#!!!Make sure you do not initialize Miniconda as this will break you loading other Anaconda modules on the cluster!!!

./Miniconda3-latest-Linux-x86_64.sh

#Make a modules folder for yourself
mkdir /home/$USER/modules

#Make a module file for the install you just made
nano /home/$USER/modules/minconda

#Put the correct stuff inside said module file - i.e. paste the below into the file - if you are not using the miniconda subdirectory in your data folder - make sure you change the below - also note that this file is in TCL not bash - so environmental variable handing is different from the normal $USER for username:

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------

```
#%Module10.2#####################################################################
##
## User Data Directory Miniconda module file
##
#  Updated by JKWMoore 10/12/2020
#
#
################################################################################


proc ModulesHelp { } {
        global version

        puts stderr "Makes a user's personal install of Miniconda in their data directory available."
}

module-whatis   "Makes a user's personal install of Miniconda in their data directory available."

# module variables
#
 set MINICONDA_DIR     /data/$env(USER)/miniconda/bin

prepend-path PATH $MINICONDA_DIR
```

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------

[...]

PR #1142, opened 2021-02-01
Add Miniconda installation documentation to the Python pages

jkwmoore:
WRT  https://github.com/rcgsheffield/sheffield_hpc/issues/1124

Might be some weirdness with the hr lines.

### DIFF
--- a/bessemer/software/apps/python.rst
+++ b/bessemer/software/apps/python.rst
@@ -25,7 +25,7 @@ Using conda Python
 After connecting to Bessemer, start an interactive session
 with the ``srun --pty bash -i command``.
 
-Anaconda Python can be loaded with one of the following:
+Anaconda Python can be loaded with one of the following: ::
 
     module load Anaconda3/2019.07
     module load Anaconda3/5.3.0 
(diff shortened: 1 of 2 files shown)


bessemer/software/apps/python.rst
```diff
@@ -141,3 +141,82 @@ Then submit this to Slurm by running:
...
+
+    chmod +x Miniconda3-latest-Linux-x86_64.sh
+
+Run the installer but make sure to choose your home directory and subdirectory i.e. /home/$USER/miniconda (n.b. don't make this directory prior to intalling as it makes the directory itself when you supply it with a directory path.)
+
+``Ensure you do not initialize Miniconda as this will break you loading other Anaconda modules on the cluster!`` ::
+
+    ./Miniconda3-latest-Linux-x86_64.sh
+    
+Make a modules folder for yourself: ::
```
willfurnass, line 176: I think users should activate conda using `source /path/to/miniconda3/etc/profile.d/conda.sh` rather than create their own modulefiles - it's quite a bit simpler.  Recommended here: https://conda.io/projects/conda/en/latest/release-notes.html?highlight=conda.sh#recommended-change-to-enable-conda-in-your-shell

bessemer/software/apps/python.rst
```diff
@@ -141,3 +141,82 @@ Then submit this to Slurm by running:
...
+
+    module use /home/$USER/modules
+
+To skip doing the above everytime you login, you can add this line to the .bashrc file in your home directory.
+
+You can load the module whenever you need it with (assuming you named the module file miniconda): ::
+
+    module load miniconda
+    
+``Avoid loading any two modules of the same type at the same time - i.e. do not load our anaconda modules and miniconda at the same time.``
```
willfurnass, line 220: If we switch to advocating for sourcing `conda.sh` then this could be amended to read something like: "don't try using the centrally-installed conda and your own miniconda install at the same time".

sharc/software/apps/python.rst
```diff
@@ -164,6 +165,87 @@ Currently, this channel provides Conda packages for:
...
+
+    chmod +x Miniconda3-latest-Linux-x86_64.sh
+
+Run the installer but make sure to choose your data directory and subdirectory i.e. /data/$USER/miniconda (n.b. don't make this directory prior to intalling as it makes the directory itself when you supply it with a directory path.)
+
+``Ensure you do not initialize Miniconda as this will break you loading other Anaconda modules on the cluster!`` ::
+
+    ./Miniconda3-latest-Linux-x86_64.sh
+    
+Make a modules folder for yourself: ::
```
willfurnass, line 201: See comments for Bessemer page.

bessemer/software/apps/python.rst
```diff
@@ -141,3 +141,82 @@ Then submit this to Slurm by running:
 
    sbatch myscript.slurm
 
+
+
+Installing a personal copy of miniconda
```
willfurnass, line 146: Could add a note on why people might want to install their own miniconda (IIRC to get Python >3.6 and a non-ancient `conda` tool).

willfurnass: Might also want to add something to the Troubleshooting/FAQ page on this.

willfurnass (changes requested):

bessemer/software/apps/python.rst
jkwmoore: I'm not sure that we should encourage the use of individual package loading / sourcing methods over module making?

We can also enforce conflict resolution with the modules (though I didn't add this in the example.)

I'd prefer if people were making modules as a SOP for any installation of any kind - something to discuss a bit further I think.

bessemer/software/apps/python.rst
jkwmoore: It might make sense to amend this to have a conflict with the system wide one?

bessemer/software/apps/python.rst
jkwmoore: Ah yeah I hadn't really thought about explaining why you might want to do this - I'll make some amendments and add something to the FAQ in due course.

bessemer/software/apps/python.rst
willfurnass: I'm not sure we're going to agree on this.  What do others think?  @davidwilby @desmfryan

davidwilby: I'm not sure I can add anything here but I'd be keen to hear the pros and cons of both approaches.

willfurnass: I personally think there's a lot less involved in sourcing a `conda.sh` file and it's more in keeping with documentation elsewhere on conda activation/usage, negating the benefits of creating a modulefile to avoid conflicts with the central conda install.

jkwmoore: Yeah I think I am more keen on scoping the entire HPC cluster and all software with a SOP for all rather than fragmenting / scoping for each software package with its own importing method (even if it is arguably easier.)

I don't feel too strongly about modules enough to insist but I'm certainly on the side of modules as they are the way we already do things on the cluster and will work for everything (with exception to the current issues with sourcing bash e.g. OpenFOAM) and we can enforce conflicts.

But hey, we can document both methods after all! :)

jkwmoore: WRT making a SOP - I have been putting this together: https://github.com/jkwmoore/sheffield_hpc/tree/Add-generalised-software-install-instructions 

Nowhere near done yet.

willfurnass:
No longer need to note that centrally installed Conda limited to Python <=3.6 now that central Conda upgraded (today).

bessemer/software/apps/python.rst
jkwmoore: We can also add a conflict with the conda modules if we use the module approach.

PR #1145, opened 2021-02-04
Add the right hand CSS

jkwmoore:
Probably needs tweaking a bit more to be themed nicely but a nice POC.


jkwmoore:
This addresses some of the annoying behaviour with very long menus resulting in a lot of scrolling if going between cluster software pages e.g.

image

vs

image

willfurnass:
I know little about UX so suggest I'm not one of the reviewers here :)

PR #1147, opened 2021-02-04
Add notes on how to tell conda to create envs in a different location

willfurnass:
@bobturneruk - mind having a look at this?  Would be grateful if you could check this works and that both new envs *and* the package cache are updated in the new location after putting this into effect.


bobturneruk:
Will do. Probably can't test it today, but hopefully tomorrow.

willfurnass:
I *think* this approach should work with interactive jobs, batch jobs and JupyterHub jobs.

PR #1171, opened 2021-03-05
add note for connection to other non-hpc servers 

anthonybrookfield:
only being possible if aggreed with infosec


willfurnass (approved):

willfurnass:
So not to be merged yet?

anthonybrookfield:
ah - no, I see how that reads now...  github wanted to split the commit line.... it's mergeable

PR #1202, opened 2021-04-29
Update to include scheduler/MPI aware containers

jkwmoore:
Updates to clarify the available methods for dealing with this, minor corrections and again trailing white spaces are gone.

### DIFF
--- a/sharc/software/apps/singularity.rst
+++ b/sharc/software/apps/singularity.rst
@@ -161,11 +161,36 @@ Automatic Mounting of ShARC Filestore Inside Images
 
 When running Singularity containers on the cluster,
 the paths ``/fastdata``, ``/data``, ``/home``, ``/scratch``, ``/shared`` and ``/tmp`` are
-automatically *bind-mounted* (exposed) from the *host* operating system to your container 
-e.g. the cluster's ``/fastdata`` directory will be automatically visible within a container started on the cluster 
+automatically *bind-mounted* (exposed) from the *host* operating system to your container
+e.g. the cluster's ``/fastdata`` directory will be automatically visible within a container started on the cluster
 without that directory being explicitly created when the corresponding Singularity image was built.
 
-If 
+Considerations for MPI or Scheduler aware containers
+----------------------------------------------------
+
+When using the batch scheduler system care has to be taken with MPI jobs or containers which are scheduler aware. For the former, in order to use the Intel OmniPath Architecture with MPI the appropriate drivers must be available in the container or the job may revert to using the gigabit ethernet or crash.
+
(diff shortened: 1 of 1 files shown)


desmfryan (commented):
James

Looks good.

The only minor niggle is:
e.g. the cluster's ``/fastdata`` directory will be automatically visible within a container started on the cluster

I think I might say something like:
e.g. the cluster's directories will be automatically visible within a container started on the cluster

so that it is not interpreted as /fastdata only....despite your earlier wording.

Cheers,

Des

sharc/software/apps/singularity.rst
```diff
@@ -161,11 +161,36 @@ Automatic Mounting of ShARC Filestore Inside Images
...
 When running Singularity containers on the cluster,
 the paths ``/fastdata``, ``/data``, ``/home``, ``/scratch``, ``/shared`` and ``/tmp`` are
-automatically *bind-mounted* (exposed) from the *host* operating system to your container 
-e.g. the cluster's ``/fastdata`` directory will be automatically visible within a container started on the cluster 
+automatically *bind-mounted* (exposed) from the *host* operating system to your container,
+i.e. the cluster's ordinary filestores will be automatically visible within a container started on the cluster
 without that directory being explicitly created when the corresponding Singularity image was built.
 
-If 
+Considerations for MPI or Scheduler aware containers
```
willfurnass, line 168: Do we want something v. short and simple like the following?

```
MPI and Singularity
-------------------

If you try running MPI within a Singularity container on a single node
the MPI implementation may auto-detect that its processes are running under the SGE job scheduler 
and will try to read a SGE file containing information on the CPU cores SGE has allocateed to that job.
This will fail unless you run your container with a ``--bind $PE_HOSTFILE:$PE_HOSTFILE:ro`` argument e.g.: ::

    singularity exec --bind $PE_HOSTFILE:$PE_HOSTFILE:ro /usr/local/packages/singularity/images/example.simg /home/$USER/my_script.sh

For a more complete guide to using MPI with Singularity (inc. multi-node jobs) `see the Singularity project's documentation <https://sylabs.io/guides/3.7/user-guide/mpi.html>`__.
```

willfurnass (changes requested):

sharc/software/apps/singularity.rst
jkwmoore: Yes, I think being prescriptive with ` --bind $PE_HOSTFILE:$PE_HOSTFILE:ro` is probably for the best - will amend this section shortly.

jkwmoore: Just addressed this in 1911fdcb83ce251e6dbffd63dec0bea049fb8647

willfurnass (approved):

PR #1207, opened 2021-05-17
Update old and insecure Python dependencies

willfurnass:


jkwmoore:
This has passed the auto test - has it had any affect / likely to have any effect on the generated docs themselves (GUI display etc...) anyone had a look?

jkwmoore (commented):
Nothing in here looks like it will affect the generated document GUI display tbh.

willfurnass:
Think the greatest risk is from the Jinja major version bump.  Will review shortly.

willfurnass:
Think @jkwmoore  + dependabot have addressed the related issues, so closing.

PR #1208, opened 2021-05-18
Adding ANSYS Pathing reference info

jkwmoore:
Depends on https://github.com/rcgsheffield/sheffield_hpc/pull/1204 - will merge this one as soon as that one is merged.


jkwmoore (commented):
Assessed the files through prior to merge - seems to be showing duplicated commits to the previous merge which shouldn't be showing as changes though...

PR #1213, opened 2021-05-19
Adds 2 sections of CSS to the custom.css file to fix table display.

jkwmoore:
Fixed as documented here: https://github.com/readthedocs/sphinx_rtd_theme/issues/117

This should fix this annoying issue elsewhere - not just on this location https://docs.hpc.shef.ac.uk/en/latest/hpc/scheduler/submit.html

Also some corrections to tables to force spacing with |br| . Anyone know if there are any other tables that might want a similar treatment?

See the current tables:

image

Versus fixed with this commit:

image


willfurnass:
Nice not having to scroll sideways with large tables but it does increase the width of the first n-1 cols in an odd way - can this be prevented?

jkwmoore:
> 
> 
> Nice not having to scroll sideways with large tables but it does increase the width of the first n-1 cols in an odd way - can this be prevented?

Yes but the solution may amuse you and it does not readily port itself to other tables. See my next commit in figures few.

desmfryan (approved):
Hi James

Looks really good......no suggestions/amendments.

Cheers,

Des

PR #1244, opened 2021-07-13
Updating the "Getting and Account" page to reflect the HPC Driving License requirements

jkwmoore:
-Majorly reworded.
-Reworked in details from https://www.sheffield.ac.uk/it-services/research/hpc-facilities
-Adjusted styling to highlight the warnings.

I'm not 100% happy with this yet as the warning not to share account details seems a little out of place.

### DIFF
--- a/hpc/accounts.rst
+++ b/hpc/accounts.rst
@@ -3,34 +3,64 @@
 Getting an Account
 ==================
 
-Before you can start using the clusters you need to register for an HPC account.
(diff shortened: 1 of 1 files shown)


jkwmoore:
I've reworded most of this to refer to getting HPC access instead of getting a HPC account on the basis of users getting confused about whether the HPC access needs a different username and password (which is an occasional ticket).

We may want to rename the page to 'Getting HPC access' if we stick to the changes above rather than reverting to 'Getting a HPC account' throughout.

PR #1248, opened 2021-08-05
JADE 2: add access form and remove warning that not yet live

willfurnass:
@twinkarma: happy with this?  Anything we can add at this stage re docs and specs?


willfurnass:
@twinkarma We might also want to put something on that page re who to contact if one has questions when filling out the form.

PR #1256, opened 2021-08-31
Gaussian stub page update.

jkwmoore:
This should fix the redirect (does in Chromium for me.)
Also adds a link if the redirect does not work.


willfurnass:
Cheers!

PR #1272, opened 2021-09-09
Implement .readthedocs.yaml fix for RTD build process.

jkwmoore:
This is adding the required .readthedocs.yaml files as well as the split requirements files in order to get setuptools pinned to the working version.

Documented in the readme file. May need rewording, or reworking in the readme file.


.readthedocs.yaml
```diff
@@ -0,0 +1,19 @@
...
+sphinx:
+   builder: html
+   configuration: conf.py
+   fail_on_warning: true
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+   version: "3.7"
+   install:
+   - requirements: ./setuptoolsrequirements.txt
```
willfurnass, line 18: Suggest adding a comment that need to downgrade setuptools before installing other dependencies until there's a new release of sphinx-bootstrap-theme with a fix for https://github.com/ryan-roemer/sphinx-bootstrap-theme/issues/216.

setuptoolsrequirements.txt
```diff
@@ -0,0 +1 @@
+setuptools==57.5.0
```
willfurnass, line 1: `setuptools <=58` more clearly shows intent?

willfurnass (commented):
Looks reasonable.  

Do you know which branches this config will apply to?  The corresponding settings in the UI are global.  I'm wondering if merging this PR would break builds for our `legacy-iceberg` branch unless we cherry-pick this commit onto the corresponding Iceberg-specific branch.

.readthedocs.yaml
jkwmoore: Just amended this.

setuptoolsrequirements.txt
jkwmoore: And amended to this also.

jkwmoore:
> Looks reasonable.
> 
> Do you know which branches this config will apply to? The corresponding settings in the UI are global. I'm wondering if merging this PR would break builds for our `legacy-iceberg` branch unless we cherry-pick this commit onto the corresponding Iceberg-specific branch.

AFAIK the yaml will only apply to the branches the file is actually in so we will probably want to port the files into the iceberg legacy docs as well or the build would also fail the next time it ran.

PR #1273, opened 2021-09-10
Update pull-request.yml

jkwmoore:
Adding the ability to run actions via manual trigger and artifact upload from each Python build process.

Draft for now but if we do wish to enable we may want to reduce the 90 day artifact retention to 30 ,14 days etc...


willfurnass:
What sorts of uses are you thinking of for this?

jkwmoore:
> What sorts of uses are you thinking of for this?

Mainly so we can get at the generated files and check display is functional - particularly for dependency or other automated PRs. You can also get an idea of the folder size but that isn't exactly a great benefit.

willfurnass:
Seems to make sense to introduce creation of build artefacts if/when we definitely need it as IIRC there are reasonably low limits for  artefact storage per GH Org.

PR #1274, opened 2021-09-13
GDAL additional driver support + docs

jkwmoore:
GPKG and SQLite driver support added for the 3.x.x versions of GDAL via the inclusion of the SQLite module and specification of the SQLite module root directory via the ``./configure`` step.


willfurnass:
Think this is waiting on our link-checking GHA, which _should_ only be triggered periodically, not by PRs.  Odd.

willfurnass:
@jkwmoore I've updated the branch protection rules for `master` so now the `test_build_docs` Workflow (with just Python 3.7) needs to pass before a PR can be merged.

PR #1297, opened 2021-09-17
Monthly is probably a nicer frequency.

jkwmoore:
Monthly is probably a nicer frequency.


jkwmoore:
We can also adjust the number of PRs it is allowed to open though setting this high could cause a similar deluge on a monthly basis.

willfurnass (approved):

jkwmoore:
I'll whack merge on this as soon as I check with @desmfryan  etc... are happy / don't want to neuter it further.

PR #1301, opened 2021-09-21
Add scheduler reference info to reference section

jkwmoore:
Added a bunch on the SLURM commands and generally  refactored things into the correct section.


jkwmoore (commented):
Gave it a quick once over again and there's nothing obviously wrong. We could also re-write the SGE command pages as I did leave these effectively unchanged, just moved.

PR #1303, opened 2021-09-23
Initial-page for Intel MPI

jkwmoore:
Should be good to go, barring the odd counting by Intel MPI. Used for Oasys LS-DYNA for which further documentation is coming.

### DIFF
--- a/sharc/software/modulefiles/mpi/impi/2019.9.304/binary
+++ b/sharc/software/modulefiles/mpi/impi/2019.9.304/binary
@@ -1,0 +1,35 @@
+#%Module1.0#####################################################################
+##
+## Intel-MPI version 2019.9.304 module file
+##
+
+## Module file logging
+source /usr/local/etc/module_logging.tcl
+##
+
+proc ModulesHelp { } {
+    global openmpiversion
+
+    puts stderr "   Adds `Intel-MPI version $intelmpiversion' to your PATH environment variable and necessary libraries"
+}
+
+set intelmpiversion   2019.9.304
+set intelmpiroot      /usr/local/packages/mpi/impi/2019.9.304/binary/
+
+module-whatis   "loads the necessary `Intel-MPI version $intelmpiversion' library paths"
+
+# Prepend paths generated through the use of env2 - see: https://sourceforge.net/projects/env2/
+# Also see: https://software.intel.com/content/www/us/en/develop/articles/using-environment-modules-with-the-intel-development-tools.html
+
+prepend-path PATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/bin/intel64:/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/bin:/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/libfabric/bin:/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/bin};
+prepend-path MANPATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/man};
+prepend-path LIBRARY_PATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/libfabric/lib};
+prepend-path CLASSPATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/lib/mpi.jar};
+prepend-path I_MPI_ROOT {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi};
+prepend-path INTEL_LICENSE_FILE {/usr/local/packages/dev/intel-ps-xe-ce/license.lic};
+prepend-path FI_PROVIDER_PATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/libfabric/lib/prov:/usr/lib64/libfabric};
+prepend-path LD_LIBRARY_PATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/compiler/lib/intel64_lin:/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/libfabric/lib:/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/lib/release:/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/lib};
+prepend-path NLSPATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/compiler/lib/intel64/locale/%l_%t/%N};
+
+# Set the Infini/omnipath fabric
+setenv I_MPI_FABRICS shm:ofi
--- a/sharc/software/parallel/intelmpi.rst
+++ b/sharc/software/parallel/intelmpi.rst
@@ -1,0 +1,216 @@
+.. _intelmpi_sharc:
+
+Intel MPI 
+=======================
+
+.. sidebar:: Intel MPI 
+
+   :Versions: 2019.9.304, 2018.5.288
+   :URL: https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/mpi-library.html
+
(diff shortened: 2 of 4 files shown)


sharc/software/parallel/intelmpi.rst
```diff
@@ -0,0 +1,216 @@
...
+   #$ -m abe
+   #$ -l h_rt=00:10:00
+   #$ -l rmem=1G
+   #$ -pe mpi 2
+   #$ -j yes
+   #$ -N intelmpi-test
+
+   module load mpi/impi/2018.5.288/binary
+
+   MACHINEFILE="machinefile.$JOB_ID"
```
willfurnass, line 57: Does `MACHINEFILE` always need to be set when using these versions of Intel MPI?

sharc/software/parallel/intelmpi.rst
```diff
@@ -0,0 +1,216 @@
...
+
+   for host in `cat $PE_HOSTFILE | awk '{print $1}'`; do
```
willfurnass, line 59: Concise expression (if it's useful :))
```sh
MACHINELIST="$(awk '{for (i=0; i<$2; i++) {print $1}}' test.txt $PE_HOSTFILE | paste -sd:)"
```

sharc/software/parallel/intelmpi.rst
willfurnass: And if so does it need to be `export`ed?

sharc/software/parallel/intelmpi.rst
```diff
@@ -0,0 +1,216 @@
...
+These are primarily for administrators of the system.
+
+Version 2019.9.304
+^^^^^^^^^^^^^^^^^^
+
+This version was installed using the CLI installer found in the protected media directory for Intel MPI. 
+Full installation was chosen following by using the process 
+`described by Intel <https://software.intel.com/content/www/us/en/develop/articles/using-environment-modules-with-the-intel-development-tools.html>`_ 
+using the `env2 <https://sourceforge.net/projects/env2/>`_ utility to 
+generate the module files.
```
willfurnass, line 187: Could be good to commit created modules to this repo.

willfurnass (commented):
Made a couple of suggestions.

sharc/software/parallel/intelmpi.rst
jkwmoore: No it doesn't need to be set, this should only be dumping the file out (for info) and not getting used by the mpirun at all although you can of course specify this with further mpirun arguments.

Even in the case where you supply it to mpirun as a path to a file it should not need exporting (if you are supplying it with the `-hostfile <hostfile>` or `-f <hostfile>`

See: https://software.intel.com/content/www/us/en/develop/documentation/mpi-developer-reference-linux/top/command-reference/mpiexec-hydra/global-hydra-options.html

https://software.intel.com/content/www/us/en/develop/documentation/mpi-developer-reference-linux/top/command-reference/mpirun.html (Which is explaining: mpirun args = mpiexec.hydra options as described in the mpiexec.hydra
section. )

sharc/software/parallel/intelmpi.rst
jkwmoore: Certainly happy to amend this.

sharc/software/parallel/intelmpi.rst
jkwmoore: These should now be available.

sharc/software/modulefiles/mpi/impi/2019.9.304/binary
```diff
@@ -0,0 +1,35 @@
...
+prepend-path MANPATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/man};
+prepend-path LIBRARY_PATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/libfabric/lib};
+prepend-path CLASSPATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/lib/mpi.jar};
+prepend-path I_MPI_ROOT {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi};
+prepend-path INTEL_LICENSE_FILE {/usr/local/packages/dev/intel-ps-xe-ce/license.lic};
+prepend-path FI_PROVIDER_PATH {/usr/local/packages/mpi/impi/2019.9.304/binary/compilers_and_libraries_2020.4.304/linux/mpi/intel64/libfabric/lib/prov:/usr/lib64/libfabric};
```
willfurnass, line 30: Should `/usr/lib64/libfabric` be in these modulefiles?

sharc/software/parallel/intelmpi.rst
willfurnass: Okay, was wondering if either version was too old to be reliably SGE-aware.

willfurnass: If wanting to print info on rank distribution between nodes then would `cat $PE_HOSTFILE` suffice?

sharc/software/parallel/intelmpi.rst
```diff
@@ -0,0 +1,213 @@
...
+         echo $host >> $MACHINEFILE
+      done
+   done
+
+   echo -e " MACHINE FILE\n"
+   echo $MACHINEFILE
+
+   MACHINELIST="$(awk '{for (i=0; i<$2; i++) {print $1}}' test.txt $PE_HOSTFILE | paste -sd:)"
+
+   mpirun -print-rank-map -np $NSLOTS IMB-MPI1
```
willfurnass, line 72: Is `-np` required or will Hydra figure out the number of available slots automatically?

sharc/software/parallel/intelmpi.rst
willfurnass: Or just `-print-rank-map`?

sharc/software/parallel/intelmpi.rst
jkwmoore: >If wanting to print info on rank distribution between nodes then would cat $PE_HOSTFILE suffice?

Yes and no, that will give what SGE thinks hydra should be doing where as forcing via the `-print-rank-map` argument makes hydra tell us what it is doing. (I was debugging with both for this reason but we can omit it entirely in an example.)

>Is -np required or will Hydra figure out the number of available slots automatically?

Have not tested without -np but I suspect it would indeed sort itself out.

sharc/software/modulefiles/mpi/impi/2019.9.304/binary
jkwmoore: I suppose that depends on whether the base system config for users has already added this. But during testing the effect of  `setenv I_MPI_FABRICS shm:ofi` was very clear so it would appear to be using the Omni path networking with the current config.

sharc/software/parallel/intelmpi.rst
jkwmoore: >If wanting to print info on rank distribution between nodes then would cat $PE_HOSTFILE suffice?

Yes and no, that will give what SGE thinks hydra should be doing where as forcing via the `-print-rank-map` argument makes hydra tell us what it is doing. (I was debugging with both for this reason but we can omit it entirely in an example.)

willfurnass: I think any scripting / env vars / parameters you include in the examples will be copy and pasted by many users unless it's very clear that they're for debugging or illustrative purposes only

jkwmoore: To be cheeky, I'd quite like it if they did copy these as / when we get a support call I'd  prefer to see this info.

willfurnass: I like the idea but think this should be as minimal as poss e.g.

```
# Show which nodes you have been allocated CPU cores on
cat $PE_HOSTFILE
```

and this should be consistent over all MPI docs for ShARC (w/ an equivalent for Bessemer MPI implementations?)

jkwmoore: I'm not sure OpenMPI and Intel MPI /hydra will have the same arguments for debugging but I'm not overly attached to keeping the debug arguments so just grabbing the PE_HOSTFILE is fine.

Although I think we need to refactor the Bessemer page given the lack of parity and some errors https://docs.hpc.shef.ac.uk/en/latest/bessemer/software/parallel/impi.html

sharc/software/parallel/intelmpi.rst
jkwmoore: Amended this with:

`MACHINELIST="$(awk '{for (i=0; i<$2; i++) {print $1}}' $PE_HOSTFILE | paste -sd:)"`

@willfurnass guessing the `test.txt` was not needed?

sharc/software/parallel/intelmpi.rst
jkwmoore: Amended this now to use indicative comments and `cat $PE_HOSTFILE`. Shortly going to be amending the Bessemer page for parity.

sharc/software/parallel/intelmpi.rst
jkwmoore: Verified that this does indeed perform ok without the -np at least for `mpi/impi/2018.5.288/binary` so one would assume this is the case for 2019 also.

PR #1309, opened 2021-09-30
Add-global-MFA-notification

jkwmoore:
This is a draft pending confirmation of date and wording. This can be merged later (unified with the rest of the comms i.e. MOTD and HPC email.)

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -24,3 +24,8 @@
   }
 
 }
+
+    .underline-bold {
+    text-decoration: underline;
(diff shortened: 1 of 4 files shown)


desmfryan:
James

yes this looks very good...almost as if I wrote it ;) I think Anthony is keen to put in something about the HPC users who use certs for access. I am still of the mind that this might trigger a whole raft of.....'we want to use this approach' and then of course inexperienced users getting into all sorts of difficulty.

Cheers,

Des

PR #1318, opened 2021-10-12
Refactoring Tier2 HPC page

jkwmoore:
Reorganises things into a master page with short summaries of external tier 2 resources and moves the existing affiliated systems to a sub-directory.


global.rst
```diff
@@ -11,5 +11,11 @@
 
     <br/>
 
+.. |hide-external-link-icons| raw:: html
```
jkwmoore, line 14: When called is inserted into the page and gets rid of the external link icon from showing up (useful when inserting hyperlinked images.)

jkwmoore: @willfurnass @desmfryan  just so you are aware I am planning on inserting this into global.

PR #1347, opened 2021-11-11
Amend ShARC Conda instructions to include moving the pkg cache 

jkwmoore:
Adjust the paths, adding pkg_dirs entry info and mkdir commands.


jkwmoore:
image

willfurnass (approved):

PR #1351, opened 2021-11-15
Add a what is HPC page.

jkwmoore:
May need a bit of tweaking or maybe some re-organisation but  looks like: 

image


desmfryan (commented):
Hi James

Looks great......my comments are minor.

1) Intro
I think I would get rid of the aircraft analogy and stick with the formal definition you have. I would suggest that in addition to HPC meaning computer or computers you address the other term 'HPC Clusters'
2) login nodes....the definition seems to be endorsing 'doing stuff' on the logins so I would suggest that here we say 'don't do anything on the logins, launch a qrsh, qrshx/qsh only' Then we will avoid the users trying to module load on the login and triggering calls......'cant run stuff'
3) When should I not use HPC?
Capitalise the word 'not'
4) For Training Purposes
Replace 'Sheffield University' with 'The University of Sheffield'
5) For non-legitimate or non-research purposes
I wonder should we include 'account sharing' here as non-legit?

But all looks great :)

Des

jkwmoore:
1. Yes I am not sure that the analogy for aircraft quite lands 😉 ? But in general I think we might need some form of analogy to explain that HPC is not exclusively tied to clusters even if that is primarily  where HPC occurs.

2. I think we should amend this to be clear that "interact with cluster" does not include running software although in the bullet point although this is covered in more detail in its dedicated paragraph. The below perhaps?

> login nodes (also known as headnodes) where users login, edit, upload / download files but should not run any intensive programs.

3. I can certainly capitalise it but do we prefer styling bold and underlining? I'll have a tinker with it locally and see what it looks like and upload some images.

4. Yep, will looks at amending that through the page.

5. Good catch I think that it is worth mentioning!

jkwmoore:
I think this looks alright: 
image

PR #1354, opened 2021-11-22
Adding further info / clarifying software Modules page.

jkwmoore:
Add further info for an alias and move searching for modules to its own section.


willfurnass:
I'm normally against suggesting aliases to users but that's such a useful one!

jkwmoore:
My hope is that people who know what an alias is can just drop it in. But people who don't won't know to just drop it in / will google aliases before blindly setting it. Can certainly amend in future if we find many people nuking their  bashrc files but fingers crossed we'll be ok.

PR #1355, opened 2021-11-22
Updates for external libraries info for R on Bessemer.

jkwmoore:
Link to module search method and explanation of toolchain matching added.


willfurnass:
Good idea!

PR #1359, opened 2021-11-26
Refactor jobs/sessions section

jkwmoore:
Hopefully this is going to make the background info and cluster / scheduler specific differences very clear.

Have made a lot of changes.


jkwmoore:
Pending an update to the readme file.

jkwmoore:
This should now be good to merge with the updated explanation to the readme (important files / folders.)

desmfryan (commented):
Hi James

1) Bessemer ls-dyna 'While using a X11 GUI forwarding supported SSH client, an interactive session can be started on ShARC with the srun --pty bash -i command which supports graphical applications' 
- Typo ShARC needs to be replaced with Bessemer.
- srun --pty bash -i needs to be removed since you already deal with launching an interactive session earlier in the section (a link to)

Just looking at qsh-vis onwards (tomorrow).

Des

hpc/scheduler/index.rst
```diff
@@ -10,20 +10,581 @@ Using interactive sessions / submitting batch jobs
 
     ./*
```
jkwmoore, line 11: We may wish to manually specify the TOC here to avoid the weird order with DRMAA in the menu.

PR #1362, opened 2021-11-29
Fix to: Fluent,LS-DYNA,Mechanical on ShARC.

jkwmoore:


jkwmoore:
Strictly speaking the ``-V`` should never have been needed and since we are no longer relying on ANSYS products auto-detecting SGE fully the option should no longer be needed.

Given that the arg will cause issues if people submit from a running interactive session it is probably better to take this out of the examples but leave it sign posted as above.

PR #1372, opened 2021-12-01
Make troubleshooting FAQ searchable.

jkwmoore:
Adjusted to use a HTML import, a small amount of javascript and added some keywords to each item container using a <div class="hidden">

When using the search items not matching the search are set invisible. 

See: https://streamable.com/z8k5w8


jkwmoore:
I've had a gander at whether doing this breaks the main search which it does not - but it will only be able to link to the entire troubleshooting page rather than specific subsections directly. See:

image

Versus current:

image


Probably due to the item-container div which is needed to hide things so this is likely mutually exclusive. The hidden divs are also searchable in the main search which is ideal. So I am inclined to merge.

PR #1386, opened 2021-12-23
Add HPC changelog links.

jkwmoore:


jkwmoore:
Further links may need adding elsewhere!

desmfryan (approved):
James

Looks good. The only comment is not specific to changelog just that on this page the request for software install points to github....need to bear this in mind depending on the TopDesk forms and if we go for a 'software install request form'

Des

PR #1435, opened 2022-02-18
Add sinfo and squeue examples for dcs-gpu partitons

ptheywood:
Adds example usage of `sinfo` and `squeue` to query GPU node information for the DCS GPU nodes, via the `dcs-gpu` and `dcs-gpu-test` partiton options.

Includes example output of these commands, although this could be removed.

If this is deemed useful, I can replicate this for the `dcs-acad` partitions, and the public GPU partition(s) too? (or generalise  to one location and cross ref?)


bessemer/groupnodes/dcs-gpu-nodes.rst
```diff
@@ -111,3 +111,48 @@ Resource limits per job:
    These jobs can be preempted by jobs submitted to the ``dcs-gpu`` and ``dcs-gpu-test`` partitions;
    if this happens
    the preempted jobs will be stopped mid-execution and re-queued.
+
+
+Checking Queue and Node Status
+------------------------------
+
+Using the ``squeue`` and ``sinfo`` :ref:`SLURM Commands<slurm_info>` it is possible to query the status of these nodes.
```
jkwmoore, line 119: I think we are likely to be removing these scheduler pages (under each cluster tab) and setting them as redirects to the specific sections in :

https://docs.hpc.shef.ac.uk/en/latest/hpc/scheduler/index.html#job-submission-control-on-sharc
https://docs.hpc.shef.ac.uk/en/latest/hpc/scheduler/index.html#job-submission-control-on-bessemer

It might make sense to alternatively link to our reference section? https://docs.hpc.shef.ac.uk/en/latest/referenceinfo/scheduler/SLURM/Common-commands/index.html

The reference section is intended to be the quick lookup and I've been utilizing imports to get details added to the reference area as well as main pages via ./referenceinfo/imports

This area does have some imports available for squeue / sinfo which you could use but besides the link the rest of the PR looks fine.

jkwmoore:
> Adds example usage of `sinfo` and `squeue` to query GPU node information for the DCS GPU nodes, via the `dcs-gpu` and `dcs-gpu-test` partiton options.
> 
> Includes example output of these commands, although this could be removed.
> 
> If this is deemed useful, I can replicate this for the `dcs-acad` partitions, and the public GPU partition(s) too? (or generalise to one location and cross ref?)

This is great! I'd prefer a generalised solution so we could look at making a specific import to be included in multiple pages?

ptheywood:
> This is great! I'd prefer a generalised solution so we could look at making a specific import to be included in multiple pages? 

An importable section to include in multiple pages would be ideal, but I'm not sure how to achieve that with replacements (i.e. the partition names) in sphinx.
So I'm torn between duplication with specialisation so users can find the exact commands easily or simpler maintenance.

Especially for the `squeue` command which needs all partitions that can run on those nodes to be strictly correct.
`sinfo` does take multiple partitions too, but when combined with `-N` it outputs nodes for each partition they participate in (I didn't look into deduplication of this).

I'll have a think and see what I can come up with (probably next week)

jkwmoore:
I'd have to do a bit of Google to check if it is possible with imports.We may be able to replicate what I have been doing on some of my most recent software pages.

See the top of this for the substitutions: https://github.com/rcgsheffield/sheffield_hpc/commit/85bfbcfd7af8ab2fdb770effc39e6839df7fbea4

ptheywood:
There is a sphinx extension which can be used to provide optional use of substitions inside a code block, but it requires python >= 3.8 according to the docs.

https://github.com/adamtheturtle/sphinx-substitution-extensions

jkwmoore:
For posterity this appears to have worked correctly on the live documentation.

PR #1448, opened 2022-03-15
Correction to OMP_NUM_THREADS

jkwmoore:
Correction to match OMP_NUM_THREADS to the number of cores.


willfurnass:
Could recommend setting to `$NSLOTS`?

jkwmoore:
> Could recommend setting to `$NSLOTS`?

True, I was thinking about that and rewriting the PR to cover all the instances where `OMP_NUM_THREADS` occurs with SMP to use `$NSLOTS`.

I'll get round to that in a bit!

PR #1454, opened 2022-03-24
New install JAGS on Bessemer, tidy up on JAGS page for ShARC.

jkwmoore:
As above.


jkwmoore:
Checks are stuck. Recreating.

PR #1471, opened 2022-04-12
Update R&I team name

willfurnass:
We're now Research and Innovation team, not Research Computing Group.  Also updated link to more info on team.

### DIFF
--- a/bessemer/software/apps/ansys/ls-dyna.rst
+++ b/bessemer/software/apps/ansys/ls-dyna.rst
@@ -145,8 +145,8 @@ Due to the limited number of licenses available if issues are encountered with r
 jobs please check the logs to see if the program is indicating an insufficient number 
 of available licenses. If this is the case, please resubmit your job until it runs.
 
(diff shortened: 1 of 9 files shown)


jkwmoore:
We also need to have a gander through / chat about the ITS wiki to sort this via grouping everyone under R&I and then tying the different services to the group (or sub-groups) too.

Was looking at this the other day and we have a bit of a mixture of different out of date groups but it might be a bit of a major refactor (and I don't want to arbirarily just start changing things on my own.)

jkwmoore (approved):
Changes LGTM but we also need to amend the help page (https://docs.hpc.shef.ac.uk/en/latest/help.html) and some other bits I suspect are floating about.

willfurnass:
> Changes LGTM but we also need to amend the help page (https://docs.hpc.shef.ac.uk/en/latest/help.html) and some other bits I suspect are floating about.

Tried to do this with the amended commit.

jkwmoore (approved):
Double approve, LGTM. Will keep in mind updating this for our future docs also.

PR #1472, opened 2022-04-13
Git on ShARC: replace 2.19.2 with 2.35.2 (CVE-2022-24765).

willfurnass:


jkwmoore (approved):
LGTM. Merging on the 22nd I take it?

Just having a quick dig around to see if we might need to do a bit of comms - I remember there were some changes between Git versions that caused some chmod weirdness with the shared areas.

willfurnass:
Keen to merge ASAP.  

The use of git on /shared areas is something we need to review anyway, so don't see that as being a blocker to merging this.  2.19.2 was only used by 4-5 people in the last year.

PR #1489, opened 2022-05-12
Add a storage selection decision tree diagram to the file stores page.

jkwmoore:
This can be further refined if desired but I'd like to merge ASAP.


willfurnass:
This is _very_ nice!  

A couple of comments: 
 - Could note that `/scratch` is node-local (important on ShARC)
 - Instead of 'does your job read or write large files' we could possibly have 'does your job read or write large files or are you not particularly concerned about file read/write performance'? (but with better wording)

Also, how do we update this, should we ever need to?  Some comments in the repo's README would be handy.

jkwmoore:
- I'll probably make the amendment for scratch being node local (and the below) on this issue's PR: https://github.com/rcgsheffield/sheffield_hpc/issues/1490

- I think we could amend this as a further section "Is file read/write performance not important?" as a further step below "Does your job read or write large files?" and direct this to /data on ShARC and /home on Bessemer (backups/snapshots being why.)

- I've got the source saved for the diagram in GDrive fortunately! I just need to migrate it to the shared drive in a sensible location or store it in the repo itself.

I'll look at sorting this on Monday as I am feeling a bit too tired to cram this all in the next 10 minutes!

willfurnass:
If we wanted something version-controllable then a Mermaid.js diagram _might_ work but would require re-making the diagram using different tooling :(

jkwmoore:
True, I might have a faff about with it and see exactly what formats it can output as it seemed quite flexible / feature rich. (Draw.io)

PR #1495, opened 2022-05-18
Add info on temporary A100 GPU nodes in Bessemer

### LINKED ISSUE #1484: Document the new A100 GPU node usage instructions:
This should probably be added under: https://docs.hpc.shef.ac.uk/en/latest/bessemer/GPUComputingBessemer.html

Ensuring that we make it clear that this is a temporary service before they get moved to their proper home.

This is pending their general accessibility after testing is complete.

willfurnass:
Pretty terse, but that's in part deliberate because users of these nodes should ideally be already familiar with Bessemer/Slurm.

NB name of the partition these A100 nodes are in wants changing to `gpu-a100-tmp` to match these docs.

Closes #1484

### DIFF
--- a/bessemer/GPUComputingBessemer.rst
+++ b/bessemer/GPUComputingBessemer.rst
@@ -168,6 +168,11 @@ Specifications per A100 node:
    AMD-compatible modules by 'unusing' the default modules area and 'using' the alternate ``eb-znver3`` area.
    Available modules can then be listed with the ``modules avail`` command."
 
+   Note that if you've compiled any software in your :ref:`home, fastdata or shared areas <filestore>`
+   using other nodes in Bessemer then this may or may not run on these AMD nodes
+   depending on the extent to which the compilation process optimised the code for Intel CPUs.
+   You may need to recompile.
+
    Many users of these nodes will want to use the 
    :ref:`Conda <python_conda_bessemer>` package manager 
    to install software; 
--- a/bessemer/GPUComputingBessemer.rst
+++ b/bessemer/GPUComputingBessemer.rst
@@ -143,7 +143,7 @@ Specifications per A100 node:
 * Chassis: Dell XE8545
 * CPUs: 48 CPU cores from 2x AMD EPYC 7413 CPUs (*AMD Milan* aka *AMD Zen 3* microarchitecture; 2.65 GHz; 128MB L3 cache per CPU)
 * RAM: 512 GB (3200 MT/s)
-* Local storage: 460 GB RAID 1 on SSDs (boot device) plus 2.6 TB RAID 0 on SSDs (:ref:`'/scratch' temporary storage<scratch_dir>`)
+* Local storage: 460 GB boot device (SSD) plus 2.88 TB :ref:`'/scratch' temporary storage<scratch_dir>` (RAID 0 on SSDs)
 * GPUs: 4x `NVIDIA Tesla A100 <https://www.nvidia.com/en-gb/data-center/a100/>`__, each with
 
   * High-bandwidth, low-latency `NVLink <https://www.nvidia.com/en-gb/design-visualization/nvlink-bridges/>`__ GPU interconnects
--- a/bessemer/GPUComputingBessemer.rst
+++ b/bessemer/GPUComputingBessemer.rst
@@ -134,7 +134,7 @@ As of May 2022, 16 addtional GPUs nodes are (temporarily) available to all users
 These feature **NVIDIA A100 GPUs, which may be quite a bit faster than the (older generation) V100 GPUs** available elsewhere in Bessemer.
 
 .. warning::
-   These nodes are being made availble to Bessemer's users on a temporary basis
+   These nodes are being made available to Bessemer's users on a temporary basis
    as they will be migrated to the University's new cluster
    when that comes online in summer 2022.
 
--- a/bessemer/GPUComputingBessemer.rst
+++ b/bessemer/GPUComputingBessemer.rst
@@ -189,29 +189,39 @@ Specifications per A100 node:
       deactivate
       . activate my_non_intel_env2
 
-**Interactive usage/access**: ::
+Interactive usage/access
+^^^^^^^^^^^^^^^^^^^^^^^^
 
- # Start an interactive session on the A100 nodes, this case with just one A100 GPU:
- srun --pty --partition=gpu-a100-tmp --gpus-per-node=1 /bin/bash -i
+Start an interactive session on the A100 nodes (in this case with just one A100 GPU): ::
+
+   srun --pty --partition=gpu-a100-tmp --gpus-per-node=1 /bin/bash -i
  
- # Activate software that has been optimised for the AMD Milan CPUs in these nodes
- module unuse /usr/local/modulefiles/live/eb/all 
- module unuse /usr/local/modulefiles/live/noeb
- module use /usr/local/modulefiles/staging/eb-znver3/all/
+Activate software that has been optimised for the AMD Milan CPUs in these nodes: ::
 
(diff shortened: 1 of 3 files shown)


bessemer/GPUComputingBessemer.rst
```diff
@@ -122,6 +122,60 @@ GPU-enabled Software
...
+   These nodes are being made availble to Bessemer's users on a temporary basis
+   as they will be migrated to the University's new cluster
+   when that comes online in summer 2022.
+
+Specifications per A100 node:
+
+* Chassis: Dell XE8545
+* CPUs: 48 CPU cores from 2x AMD EPYC 7413 CPUs (*AMD Milan* aka *AMD Zen 3* microarchitecture; 2.65 GHz; 128MB L3 cache per CPU)
+* RAM: 512 GB (3200 MT/s)
+* Local storage: 460 GB RAID 1 on SSDs (boot device) plus 2.6 TB RAID 0 on SSDs (temporary storage)
```
jkwmoore, line 144: I'm guessing this is bound to /scratch as per normal? Might be worth just amending to ``/scratch temporary storage`` if that is the case.

bessemer/GPUComputingBessemer.rst
```diff
@@ -122,6 +122,60 @@ GPU-enabled Software
...
+
+.. warning::
+
+   Much of the :ref:`existing centrally-provided software on Bessemer <bessemer-software>` won't work on
+   the AMD CPUs in these nodes as it has been heavily optimised for Intel CPUs.
+
+   Attempts to run Intel-optimised software on these nodes
+   will often result in *illegal instruction* errors.
+
+   A limited number of software packages have been provided for use on these nodes
```
jkwmoore, line 158: We might get a bit of a "What? Where?" effect here if people are not familiar with modules commands, even though the module commands are shown below. We might want to add a bit saying:

```
 "See the example below where we change our modules source from the normal Bessemer 
 software packages modules to AMD compatible modules by unusing the default modules 
 area and using the alternate ``eb-znver3`` area.
 
 Available modules can be listed with the ``modules avail`` command."
```
 
 Then delete L171-172?

willfurnass:
Added info on conda + BLAS libs, inc OpenBLAS being more performant than the Intel MKL (unless using a special trick with an older MKL).

bessemer/GPUComputingBessemer.rst
```diff
@@ -122,6 +122,91 @@ GPU-enabled Software
...
+ # List software available for use on these nodes
+ module avail
+
+**Batch job usage/access** - within your job script(s):
+
+* Ensure you have ``#SBATCH --partition=gpu-a100-tmp`` near the top of your job script
+* Below that include the three ``module unuse`` / ``module use`` lines shown above before you run any software
+
 Training materials
 ^^^^^^^^^^^^^^^^^^
```
jkwmoore, line 211:
```suggestion
------------------
```

This should bump it up into the right bit of the TOC.

bessemer/GPUComputingBessemer.rst
```diff
@@ -122,6 +122,91 @@ GPU-enabled Software
...
+   Or you can try using an older version of the Intel MKL instead,
+   where there's a special trick for enabling better performance on AMD CPUs: ::
+
+      conda create -n my_non_intel_env2 -c anaconda python numpy mkl=2019.* blas=*=*mkl
+      . activate my_non_intel_env2
+      conda env config vars set MKL_DEBUG_CPU_TYPE=5
+      deactivate
+      . activate my_non_intel_env2
+
+**Interactive usage/access**: ::
```
jkwmoore, line 192:
```suggestion
Interactive usage/access
^^^^^^^^^^^^^^^^^^^^^^^
 ::
```

We could make this part of the TOC as above.

bessemer/GPUComputingBessemer.rst
```diff
@@ -122,6 +122,91 @@ GPU-enabled Software
...
+ 
+ # Activate software that has been optimised for the AMD Milan CPUs in these nodes
+ module unuse /usr/local/modulefiles/live/eb/all 
+ module unuse /usr/local/modulefiles/live/noeb
+ module use /usr/local/modulefiles/staging/eb-znver3/all/
+
+ # List software available for use on these nodes
+ module avail
+
+**Batch job usage/access** - within your job script(s):
```
jkwmoore, line 205:
```suggestion
Batch job usage/access
^^^^^^^^^^^^^^^^^^^^^^

Within your job script(s):
```

We could also make this part of the TOC as above.

bessemer/GPUComputingBessemer.rst
```diff
@@ -122,6 +122,91 @@ GPU-enabled Software
...
+   as they will be migrated to the University's new cluster
+   when that comes online in summer 2022.
+
+Specifications per A100 node:
+
+* Chassis: Dell XE8545
+* CPUs: 48 CPU cores from 2x AMD EPYC 7413 CPUs (*AMD Milan* aka *AMD Zen 3* microarchitecture; 2.65 GHz; 128MB L3 cache per CPU)
+* RAM: 512 GB (3200 MT/s)
+* Local storage: 460 GB RAID 1 on SSDs (boot device) plus 2.6 TB RAID 0 on SSDs (:ref:`'/scratch' temporary storage<scratch_dir>`)
+* GPUs: 4x `NVIDIA Tesla A100 <https://www.nvidia.com/en-gb/data-center/a100/>`__, each with
```
ptheywood, line 147: It might be worth mentioning these are CUDA Compute Capability 8.0 devices, which should be targetted when compiling CUDA codes (with CUDA >= 11.0 where it was introduced).

For packaged cuda applications such as Tensorflow, these should be built for 80 by default for recent enough versions.

The [Bessemer CUDA docs](https://docs.hpc.shef.ac.uk/en/latest/bessemer/software/libs/cuda.html?highlight=cuda#gpu-code-generation-options) include a section on how to pass this to nvcc for volta (70) already. I'm not sure its worth editing that page given the Ampere GPUs are only in bessemer short term. If usesrs follow the gencode instructions there it does embed PTX, so code will run if followed but will not use ampere specific functionality.

For use in the temp queue which will only ever contain A100s, `-gencode=arch=compute_80,code=sm_80` would be sufficient. No need to target older arch's or embed PTX for future arch's in this case. (Next cluster docs should still recommend ptx embedding).

It might make more sense for this be addressed in a new section towrads the end, if not just stating they're CC 8.0.

willfurnass: Thanks @ptheywood.  I've augmented the general Bessemer CUDA docs page with some of the above info and have pointed at that from the new page.

bessemer/GPUComputingBessemer.rst
```diff
@@ -122,8 +122,117 @@ GPU-enabled Software
...
+-------------------------------
+
+Prior to May 2022 the Bessemer cluster only featured :ref:`one 'public' GPU node <bessemer-gpu-specs>` containing NVIDIA V100 GPUs
+(although private GPU nodes could be accessed using :ref:`preemptable jobs, for those whose workflows could tolerate preemption<preemptable_jobs_bessemer>`).
+
+As of May 2022, 16 addtional GPUs nodes are (temporarily) available to all users of Bessemer.  
+These feature **NVIDIA A100 GPUs, which may be quite a bit faster than the (older generation) V100 GPUs** available elsewhere in Bessemer.
+
+.. warning::
+   These nodes are being made availble to Bessemer's users on a temporary basis
```
jkwmoore, line 137:
```suggestion
   These nodes are being made available to Bessemer's users on a temporary basis
```

jkwmoore:
With the fix above for the :ref: link LGTM.

PR #1496, opened 2022-05-23
Public A100 nodes: state which Slurm qos to use

willfurnass:


jkwmoore (approved):
LGTM, will follow up in Slack with some related questions (but not relevant to this PR.)

PR #1512, opened 2022-07-05
VASP corrections / updates prior to Stanage.

jkwmoore:
Various small corrections to both cluster VASP documentation pages.

### DIFF
--- a/bessemer/software/apps/vasp.rst
+++ b/bessemer/software/apps/vasp.rst
@@ -15,7 +15,7 @@ The Vienna Ab initio Simulation Package (VASP) is a computer program for atomic 
 Usage
 -----
 
-VASP 5.4.1 can be activated using the module file::
+VASP 5.4.1 or 5.4.4 can be activated using one of the module load commands below: ::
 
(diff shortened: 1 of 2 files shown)


bessemer/software/apps/vasp.rst
```diff
@@ -53,8 +54,8 @@ Users are encouraged to write their own batch submission scripts. The following
     #SBATCH --mail-user=username@sheffield.ac.uk
     #SBATCH --mail-type=ALL
 
-    module load VASP/5.4.1-intel-2019b
-    mpirun -np 4 vasp_std
+    module load VASP/5.4.4-intel-2019b
+    srun --export=ALL vasp_std
```
jkwmoore, line 58: The fact that the `--export=ALL` here is required is odd and probably warrants a bit of investigation. My understanding of https://slurm.schedmd.com/srun.html#OPT_export_1 is that this should be the default behaviour but given a direct `srun` of vasp_std results in `slurmstepd: error: execve(): vasp_std: No such file or directory
 ` something ain't right.

PR #1515, opened 2022-07-13
/data areas on ShARC now backed up

willfurnass:
- [ ] Update text on `filestore` page 
- [ ] Update flowchart on `filestore` page
- [ ] Check if need to also update snapshot retention policy info?


jkwmoore:
LGTM with exception to a missing arrowhead bottom left! I've fixed that up and the next commit will have the updated XML/HTML.

jkwmoore (approved):
As mentioned, I've amended to fix the arrowhead but you might just want to double check before merge.

PR #1526, opened 2022-08-22
New version of NAG lib (nll6i285bl)

willfurnass:
And deprecate older unlicensed versions.


desmfryan:
Hi Will

Looks good to merge.

Cheers,

Des

PR #1529, opened 2022-08-23
Intel 2020a and 2020b updates.

jkwmoore:
Adds modules to each page for IMPI and compilers.


willfurnass:
LGTM.  Think need to update https://docs.hpc.shef.ac.uk/en/latest/bessemer/software/libs/imkl.html too.

jkwmoore:
Good catch, I'll add those now.

willfurnass (approved):

PR #1540, opened 2022-08-31
Update details of ShARC public big mem nodes

willfurnass:


jkwmoore (approved):
LGTM, I'll trust the numbers / type of big mem node are accurate as I've not had a look!

PR #1573, opened 2022-11-16
Updated the sheffield hpc docs to reflect the newest version of Ansys software

Lizzydapsy:
Updated the sheffield hpc docs to reflect the newest version of Ansys software

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,26 +1,26 @@
 name: Deploy site static content to Github Pages
 
 on:
-  # Runs on pushes targeting the default branch
+  # Runs on pushes targeting the default branch.
   push:
     branches: ["master"]
 
-  # Or manually from the Actions tab
+  # Or manually from the Actions tab.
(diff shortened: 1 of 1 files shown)


Lizzydapsy:
?

Lizzydapsy:
oh, great

jkwmoore:
Here's a nice list of some of the short hands that get used: https://www.freecodecamp.org/news/what-do-cryptic-github-comments-mean-9c1912bcc0a4/

Lizzydapsy:
haha, thanks for the link. I had to quickly google LGTM :)

willfurnass:
Nice work, cheers.

Lizzydapsy:
@ Des, I intentionally did not changed that because it states "The arguments -sgepe mpi-rsh -sge are not required and will cause issues if used with Fluent 21.1 or above". The statement has captured 22.1 (in the word "above"). If I change that, then it will imply that using the argument with 21.1 will not have issues, which isn't right.

PR #1578, opened 2022-11-17
Update static.yml

jkwmoore:
Fixes the Action getting stuck due to 'mv' error code.


willfurnass:
Redirect to `/en/latest` looking good.

PR #1582, opened 2022-11-18
Remove invalid LS-DYNA licence warning.

jkwmoore:
As of 2022 academic year and onward, LS-DYNA licences are now numerous and available in our base multiphysics licence.

Thus the warnings are no longer valid.


Carldkennedy (approved):
LGTM

nikishe (commented):
Good edit

PR #1585, opened 2022-11-25
Stanage skeleton docs

desmfryan:
Added skeleton docs for openfoam & vasp (pilot user software).

There are multiple entries containing (xxx TO BE CONFIRMED xxx).
Module files are not present NOTE they point to the Bessemer module files
Installation notes not correct NOTE they point to Bessemer easybuild files

Interactive session for OpenFoam points to existing docs 'Using the HPC Systems'


jkwmoore:
This PR is targeting master, not my separate branch FYI. The separate branch is called "Stanage_Branch".

Not sure you can redefine the PR target so you might need to remake.

PR #1588, opened 2022-12-07
Issue template suggestion.

jkwmoore:
As stated, feel free to tweak wording or add additional links.


jkwmoore:
Note: if desired for software installations we could use GitHub issue forms: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

But this depends on how we want to get installation requests since this would cut TD out of the loop.

PR #1589, opened 2022-12-07
Update MobaXterm instructions

Carldkennedy:


jkwmoore:
Have we tested if SCP fast mode works and what if any impact it has? IIRC it forces a more modern protocol should we should recommend it if it works and there's no performance regressions.

willfurnass:
This PR will need rebasing on `master` now that #1593 has been fixed on `master`.

jkwmoore (approved):
LGTM free to merge!

PR #1594, opened 2022-12-21
A100 GPU nodes to be removed from Bessemer and migrated to new HPC system

willfurnass:
I've removed some content that can be re-added once Stanage is online and have reworded other content to note that the A100 nodes are to very shortly transition between clusters with some downtime.


willfurnass:
Whoops, commited some extraneous files there; will fix.

willfurnass:
Extraneous files now removed from this PR.

PR #1603, opened 2023-01-05
Simplify state codes table

Carldkennedy:
@jkwmoore is this the kind of simplification you were thinking of?


jkwmoore:
I was thinking more that we acquire the same tables as the ones I linked on the Ubuntu Manpage but collapse parts of the tables so that only the most common error codes / job states are viewable without a user clicking to unfurl the whole table.

ala this https://codepen.io/andornagy/pen/gaGBZz

jkwmoore (approved):

PR #1608, opened 2023-01-05
Add auto section label extension

Carldkennedy:
https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#automatically-label-sections


Carldkennedy:
@jkwmoore Passed tests with
autosectionlabel_maxdepth <=2

jkwmoore:
If you're content that this doesn't cause any issues I don't see any reason not to merge but some further clarity on:

  'has the form {path/to/page}:{title-of-section}.' 
  
  May be needed. e.g. are these dashes replacements of white spaces? Or are they real dashes? Are there any white space underscore replacements etc...

jkwmoore (approved):
LGTM, might just be worth reviewing my comments prior to merge and adding some further explanation if needed.

Carldkennedy:
> Is there a risk that we'll end up with two separate ways of referencing a section (explicit vs implicit labelling)? Also, will implicit labels be harder to search for in the .rst source files?

It would be possible to add a prefix to all automatically generated labels - so searching would be easy enough.  
But agree it's better to be consistent. It could take a while to replace all explicit labels and references. Probably best to continue as is.

Closing

PR #1616, opened 2023-01-09
Add man usage to FAQ

Carldkennedy:


jkwmoore (changes requested):
I believe this needs `apropos` rather than `apropros`. `whilst` may also want to be `Whilst`, otherwise LGTM.

Carldkennedy:
> I believe this needs `apropos` rather than `apropros`. `whilst` may also want to be `Whilst`, otherwise LGTM.

Fixed. Thank you @jkwmoore

willfurnass:
Mostly looking good though!  Thanks for adding this.

Can also now add a label to this section and think about if/where elsewhere in the docs we want to add references to it.  The sections on SGE/Slurm might benefit from links to this as the man pages for both are very comprehensive  

Also, note that there's an issue with the man page setup on worker nodes: as soon as you load an environment module that manipulates `MANPATH` the OS-provided man pages are no longer accessible.  Should probably think about how to address that as a separate Issue...

Carldkennedy:
Thank you @willfurnass. I've made the changes you suggested. Will look for places to insert references to it.

PR #1617, opened 2023-01-11
New account file transfer with rsync FAQ

nikishe:
New Account file transfer with rsync FAQ


FAQs.rst
```diff
@@ -375,6 +375,29 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
 
 ------
 
+I Have a new account, how do I transfer data from my old account
+----------------------------------------------------------------
+To transfer data between your old account and your new account you could make use of either `SCP <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-scp-in-the-terminal>`__ or `RSYNC <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-rsync>`__. We encourage users to use RSYNC as it preserves timestamps and permisions. Follow the following workflow to carry out the transfer.
+* Log into your new username in the cluster you want to copy to and create a folder named "OldUserAccount".
```
Carldkennedy, line 381: I think you need to add a new line before the bullet point

FAQs.rst
```diff
@@ -375,6 +375,29 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
...
+I Have a new account, how do I transfer data from my old account
+----------------------------------------------------------------
+To transfer data between your old account and your new account you could make use of either `SCP <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-scp-in-the-terminal>`__ or `RSYNC <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-rsync>`__. We encourage users to use RSYNC as it preserves timestamps and permisions. Follow the following workflow to carry out the transfer.
+* Log into your new username in the cluster you want to copy to and create a folder named "OldUserAccount". 
+
+.. code-block:: console
+
+        mkdir OldUserAccount
+
+* Log into your old account and run the RSYNC command. There are two ways to run it, the first examples assumes you are copying the files to the new account on the same cluster node(e.g old account on Bessemer to new account on Bessemer), here we are only going to use the "avP" options as we dont need to compress the data.
```
Carldkennedy, line 387: * Log into your old account and run the RSYNC command. Here we show two examples. 
1) You want to copy your files to the new account on the same cluster node (e.g old account on Bessemer to new account on Bessemer), here we are only going to use the "avP" options as we dont need to compress the data.

FAQs.rst
```diff
@@ -375,6 +375,29 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
...
+
+.. code-block:: console
+
+        rsync -avP /Path/To/File_Or_Directory $Your_New_UserName@localhost:/home/$Your_New_UserName/OldUserAccount
+
+The other example assumes you are copying the files to the new account on a different cluster node(e.g old account on Bessemer to new account on Sharc/Stannage), here we are going to use the option "avzP" as we are going to transfer data over the internet, and it will be faster if it is compressed.
```
Carldkennedy, line 393: 2) You want to copy your files to the new account on a different cluster node (e.g old account on Bessemer to new account on Sharc/Stannage), here we are going to use the option "avzP" as we are going to transfer data over the internet, and it will be faster if it is compressed.

FAQs.rst
```diff
@@ -375,6 +375,29 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
 
 ------
 
+I Have a new account, how do I transfer data from my old account
+----------------------------------------------------------------
+To transfer data between your old account and your new account you could make use of either `SCP <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-scp-in-the-terminal>`__ or `RSYNC <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-rsync>`__. We encourage users to use RSYNC as it preserves timestamps and permisions. Follow the following workflow to carry out the transfer.
```
Carldkennedy, line 380: I'd also add what each of the options does and maybe about seeing full details using `man rsync`

Carldkennedy (changes requested):
Add a few comments

FAQs.rst
Carldkennedy: Also might want to add a warning about rsync push deleting files on remote if they don't exist on local?

FAQs.rst
```diff
@@ -375,6 +375,29 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
 
 ------
 
+I Have a new account, how do I transfer data from my old account
```
Carldkennedy, line 378:
```suggestion
I have a new account, how do I transfer data from my old account
```

Carldkennedy (approved):

jkwmoore (changes requested):
Should be some changes attached but otherwise LGTM.

FAQs.rst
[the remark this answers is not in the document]
jkwmoore: *when networking will affect this choice.

FAQs.rst
```diff
@@ -461,6 +461,32 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
 
 ------
 
+I have a new account, how do I transfer data from my old account
+----------------------------------------------------------------
+To transfer data between your old account and your new account you could make use of either `SCP <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-scp-in-the-terminal>`__ or `rsync <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-rsync>`__. We encourage users to use rsync as it preserves timestamps and permisions. Follow the following workflow to carry out the transfer.
```
willfurnass, line 466: scp (and sftp) can also preserve modification times; not a distinguishing feature from rsync.  However, worth mentoining that preserving modification times is typically what's required (and needs explicitly enabling).

I like the idea of recommending just one way of doing this though (using rsync) so as to keep this section concise.

FAQs.rst
```diff
@@ -461,6 +461,32 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
...
+
+* Log into your new username in the cluster you want to copy to and create a folder named "OldUserAccount". 
+
+.. code-block:: bash
+
+        mkdir OldUserAccount
+
+* Log into your old account and run the rsync command. Here we show two examples.
+
+1. You want to copy the files to the new account on the same cluster node(e.g old account on Bessemer to new account on Bessemer), here we are only going to use the "avP" options as we dont need to compress the data.
```
willfurnass, line 476: Suggest omitting mention of compression to make more concise.

Is 'avP' definitely what's required here?  Think 'a' enables archive mode, which preserves a whole bunch of things in addition to timestamps (inc ownership, which we _don't_ want here).  Also, users will want 'r' if wanting to recursively copy a directory plus contents.  Suggest running some tests and checking man page for rsync.

FAQs.rst
```diff
@@ -461,6 +461,32 @@ The RSA, ECDSA and ED25519 fingerprints for Bessemer's login nodes are: ::
 
 ------
 
+I have a new account, how do I transfer data from my old account
+----------------------------------------------------------------
+To transfer data between your old account and your new account you could make use of either `SCP <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-scp-in-the-terminal>`__ or `rsync <https://docs.hpc.shef.ac.uk/en/latest/hpc/transferring-files.html#using-rsync>`__. We encourage users to use rsync as it preserves timestamps and permisions. Follow the following workflow to carry out the transfer.
+
+* Log into your new username in the cluster you want to copy to and create a folder named "OldUserAccount".
```
willfurnass, line 468: Could say that doing this so don't accidentally overwrite files in new account home dir?

willfurnass:
Sorry, again late to the party.  Added some suggestions that could be addressed via a 2nd PR.

FAQs.rst
jkwmoore: afaik `-a `/`archive` can't and won't preserve owners unless you are the owner already, though I cannot say I have checked in recent memory. iirc `-a `should imply recursive and I do expect that it is what we want, at least from a sync for a remote ---> local transfer or vice versa.

Not sure exactly what you will get on local ----> local.

Definitely worth double checking both.

On the compression front / deeper explanation front the FAQ for this entry can simply describe the process of doing this and point out the more detailed file transfer page we have.

To decide on whether we want it for cross cluster transfers, some empirical testing needed.

nikishe: The owner after the copy is the user of the location its getting copied to , so will not create conflicts. The preserved permissions are the chmod ones so 755/777. I might be wrong on this , so please correct me if i am , but from the documentation it handles directories without the -r.
@jkwmoore I was trying to avoid redundancies , so went with the option of linking the full documentation of the rsync/scp page from the "Transferring files" section.

PR #1631, opened 2023-01-25
Votca doc edit

nikishe:
Correction of tool chains (as it was reinstalled) and a few notes on the other scientific libraries it uses.


willfurnass (approved):

willfurnass:
A couple of thoughts: would be useful to have the install script and modulefile linked to from the docs page, plus also would be useful to add a note on there being easily accessible examples.

PR #1632, opened 2023-01-27
Add git 2.39.1 to both clusters.

jkwmoore:
This is still a draft as installations are still in need of completion on Bessemer as well as the removal of the vulnerable versions.

Another PR for the changelog also required but not pending yet.

### DIFF
--- a/bessemer/software/development/git.rst
+++ b/bessemer/software/development/git.rst
@@ -1,0 +1,30 @@
+git
+===
+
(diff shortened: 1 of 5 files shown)


jkwmoore:
@Carldkennedy  / @nikishe Can one of you  give this a quick check then merge? I'm removing the older modules now.

Carldkennedy (approved):
LGTM

PR #1642, opened 2023-02-17
Unify Fluent spawn command.

jkwmoore:
As the conventional fully specified (explicitly telling Fluent where to spawn tasks) MPI spawn command for Fluent has proved the most reliable and performant even for SMP this is amended as such.


jkwmoore:
Some further work is needed to investigate whether there are good performance reasons to still elect to use Omnipath networking for MPI on the SMP jobs. (Marking draft until we have an answer.)

jkwmoore:
Use of Omnipath does not appear to help nor hinder so I am content to leave the command as is for the SMP jobs.

PR #1650, opened 2023-03-02
Add links to module pairings

Carldkennedy:
Add note pointing to table of modules to load for each version of TensorFlow

### DIFF
--- a/bessemer/software/apps/tensorflow.rst
+++ b/bessemer/software/apps/tensorflow.rst
@@ -72,6 +72,9 @@ The GPU version of TensorFlow is a distinct Pip package and
 is also dependent on CUDA and cuDNN libraries,
 making the installation procedure slightly different.
 
+.. warning::
+   You will need to ensure you load CUDA and cuDNN modules which are compatible with the version of TensorFlow used.  
+
 First request an interactive session, e.g. see :ref:`GPUInteractive_bessemer`.
 
 Then GPU version of TensorFlow can be installed by the following ::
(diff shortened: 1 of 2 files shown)


jkwmoore:
I think this might need to be a warning rather than a note, as well as being shifted to the top of the GPU install section. 

I think it would benefit people more if it is more prescriptive with the instructions about compatibility been versions and module files being *essential*.

bessemer/software/apps/tensorflow.rst
```diff
@@ -97,7 +100,7 @@ you should specify a version number when running ``pip install`` i.e. ::
    pip install tensorflow-gpu==<version_number>
 
 .. note::
-   You can find which modules to load for various versions of TensorFlow :ref:`here<tensorflow_cudnn_compat_bess>`.
+   You can find which CUDA and cuDNN module to load for various versions of TensorFlow :ref:`here<tensorflow_cudnn_compat_bess>`.
```
jkwmoore, line 103: Can we just move these into the warning?

Carldkennedy: Could do. I though it was nice to have the note right after the part showing how to use other versions. Leading into selecting the correct pairing

jkwmoore: I don't think it makes much sense to effectively start a call to action, then defer to lower down? Seems less clear / user friendly. Arguably, the first thing anyone should do before installing it is ensure the versions match so I wanted to get ahead of them trying it, then implying they should have done X first.

PR #1654, opened 2023-03-10
Fluent on Stanage: use -scheduler_tight_coupling instead of -ssh

willfurnass:
Gives us tight integration between Slurm and Fluent's bundled Intel MPI, and therefore stricter, cleaner resource management for remote Fluent tasks.

Also, remove references to 'parallel environments' as they've very much a SGE (rather than a Slurm) thing.


willfurnass:
Keen to get this merged soon as the change benefits pilot users and those running the platform.

Carldkennedy (approved):
LGTM

PR #1664, opened 2023-03-20
GPU access on SHARC: remove mention of access ever having been restricted by SGE Project

willfurnass:
Remove contradicting info.  Also remove mention of access ever having been restricted by SGE Project, as that's now very old news.


Carldkennedy (approved):
LGTM

PR #1671, opened 2023-03-29
New tab set up

nikishe:
The section "Using Python with MPI" for the Python docs might be incorrect

### DIFF
--- a/FAQs.rst
+++ b/FAQs.rst
@@ -121,17 +121,27 @@ I've loaded software but it isn't working
 
 This usually means that you are on a `login node <https://docs.hpc.shef.ac.uk/en/latest/hpc/what-is-hpc.html#login-nodes>`_. You will need to start an interactive session after which you will be able to load cluster software. 
 
-For the Bessemer cluster you will need to type the following:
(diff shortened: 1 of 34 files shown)


jkwmoore:
For now I think we will need to park this particular scheme of work (until after Stanage is live) due to the scale of work and the organisation needed to implement it with success.

When we do think we're ready to go forward with it:

A piece meal conversion for this going page by page (or software by software) into master and the live website is going to lead to a lot of confusion on which page / where to look for the currently 'valid' docs so I'd propose that for this we'd need separate tracking this work via PRs made to a non-master branch. 

That way the changes (I am expecting quite a lot of commits) can be individually implemented and reviewed via merge into the non-master branch prior to a larger PR or several large PRs into master from the 'finished' non-master branch.

nikishe:
Thats fine , I will leave the branch open for whoever wants to have a play and re-edit to the current system

jkwmoore:
> Thats fine , I will leave the branch open for whoever wants to have a play and re-edit to the current system

Cracking, thanks!

PR #1672, opened 2023-03-29
OpenFOAM on Stanage: no explicit PMIX selection needed now

willfurnass:
As PMI(X) selection now handled behind the scenes by modulefiles thanks to @jkwmoore.


jkwmoore (approved):
LGTM. Further docs on explaining dependency chain and how things like the `--mpi` option is set can go on it later if required.

willfurnass:
Agreed.  Might be best to document in 'For Admins' section per page as unlikely to be of interest to the majority of users.

PR #1678, opened 2023-04-03
SSH public key auth policy

willfurnass:
We should have one to a) encourage good practices, b) to encourage users who want to do non-standard things with SSH pub key auth to get in touch and start a conversation, and c) to dissuade users with very limited understanding of SSH pub/priv key auth from using it (which is the reason the first draft of this policy is concise and technical).


jkwmoore (approved):
Added some emphasis on the 'at creation time' aspect, otherwise LGTM.

image

nikishe (approved):
LGTM

PR #1682, opened 2023-04-04
FAQ: MPI jobs - compare srun, mpirun, mpiexec

willfurnass:
Useful to have this info in a central place with a label so it can be x-ref'ed from elsewhere (e.g. Slurm usage pages, OpenMPI pages, Intel MPI pages)


Carldkennedy (approved):
LGTM. I can see a few places to link this elsewhere in the docs :)

jkwmoore (approved):
Changed case on ShARC but otherwise LGTM -merging.

PR #1683, opened 2023-04-04
Stanage: switch to using wildcard toctrees for 4x main sw dirs

willfurnass:
Any reason why not?


jkwmoore:
> Any reason why not?

Nope. (Well, it can cause funny orders when mixing folders and files in the working directory as seen on the current software pages with the ANSYS folder mysteriously being at the top of the list compared to all the other software.)

jkwmoore (approved):

PR #1685, opened 2023-04-05
Stanage MPI builds: document compile + runtime tweaks via EasyBuild hooks

willfurnass:
I like the idea of this info being public.  Can make it public via other means if/when we publish our EB `hooks.py` file (inc explanatory comments).

### DIFF
--- a/stanagepilot/software/parallel/impi.rst
+++ b/stanagepilot/software/parallel/impi.rst
@@ -16,7 +16,7 @@ Versions
 --------
 
 You can load a specific version using one of the following: ::
-    
+
     module load impi/2019.7.217-iccifort-2020.1.217     # subset of the intel 2020a toolchain
     module load impi/2019.9.304-iccifort-2020.4.304     # subset of the intel 2020b toolchain
     module load impi/2021.2.0-intel-compilers-2021.2.0  # subset of the intel 2021a toolchain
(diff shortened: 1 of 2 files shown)


stanagepilot/software/parallel/openmpi.rst
```diff
@@ -139,7 +139,26 @@ Your output would be something like: ::
...
+     (via an environment variable - ``SLURM_MPI_TYPE=pmix_v4``) that 
+     PMIx version 4 is to be used for launching remote processes using ``srun``.
+   * Versions greater than 4.0.0 are configured at runtime to use 
+     LibFabric (OFI) for inter-process communications, which in turn is 
+     configured at runtime via environment variables to use the PSM2 OFI provider 
+     for efficient OmniPath support:
+     * ``OMPI_MCA_pml=cm``
+     * ``OMPI_MCA_mtl=ofi``
+     * ``OMPI_MCA_mtl_ofi_provider_include=psm2``
+     * ``PMIX_MCA_psec=native``
```
jkwmoore, line 161: We might just want to include why we're doing that (CentOS 7, older PSM2 lib.)

Otherwise LGTM.

jkwmoore:
```
        # With OpenMPI >=4.1.0 the PSM2 lib provided by the CentOS 7 repos is
        # too old so just use LibFabric (OFI), which the EB EasyConfigs/
        # EasyBlock enable by default with newer OpenMPIs.
 ```

File: stanagepilot/software/parallel/impi.rst [hunk 2]
```diff
@@ -137,3 +137,16 @@ This will generate output of the form:
...
     Hello world: rank 1 of 2 running on node051.pri.stanage.alces.network
+
+Installation notes
+------------------
+
+This section is primarily for administrators of the system. Intel MPI has been installed using the default Easybuild config files but with the following tweaks made via EasyBuild hooks:
+
+* Modulefiles are patched so that
+  * they instruct Slurm at runtime 
+    (via ``SLURM_MPI_TYPE=pmi2``) that
```
jkwmoore, line 148: Looks like this indentation here is causing issues, suspect this needs to be level with the bullet point?

stanagepilot/software/parallel/openmpi.rst
willfurnass: Ta; addressed that.

jkwmoore:
I'll have a quick gander with VS code now.

jkwmoore:
Last commit should fix things up as shown below:

image


image

jkwmoore (approved):
LGTM. Please whack merge on there if you're happy with the edits I made.

PR #1689, opened 2023-04-05
1686 placement of include functionality causing erroneous pages

Carldkennedy:
Moves imports to import directory


jkwmoore:
Did some very quick and nasty comments / added a new readme section, Seems good to merge.

PR #1700, opened 2023-04-13
Stange: Add CUDA 10.x gencode instructions (embed compute_70 PTX)

ptheywood:
CUDA 10.x is not aware of Ampere (Compute capability 80) GPUs, so the gencode documentation needs ammending to cover this use case.

Embedding PTX for Compute Capability 70 (Volta/GV100) should be the closest performance (It should be closer than 75/Turing due to FP64 characteristics). 
There will be some additional time on first run as the PTX is JIT'd to CC 80, but this is unavoidable if using CUDA 10.x on stange)

Default compilation (no gencodes) is fine with both, as it uses the oldest non-deprecated gencode and embeds ptx (i.e. It will run, but will not do SM 80 specific optimisations either way). 
For CUDA 10.x this will be SM 35 (2nd gen kepler, K40) IIRC, while 11.x will be SM 52 (2nd gen consumer maxwell, geforce 9xx).

For (mostly my) future reference I've put together a MWE which can be used to to test the behaviour: https://github.com/ptheywood/cuda-gencodes-demo

image

image

### DIFF
--- a/stanagepilot/software/libs/cuda.rst
+++ b/stanagepilot/software/libs/cuda.rst
@@ -74,14 +74,6 @@ An example of the use of ``nvcc`` (the CUDA compiler): ::
 
 This will compile the CUDA program contained in the file ``filename.cu``.  
 
-.. TODO: incorporate the following if install CUDA < 11 at some point.
-
-   "'nvcc filename.cu' is correct, but will only work with CUDA >= 11.0 which introdcued SM 80.
-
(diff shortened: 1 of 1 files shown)


jkwmoore:
@Carldkennedy - you may want to give this one a glance as it may relate to issues with the older CUDA / Tensorflow you were having?

@ptheywood LGTM.

Unrelated to this PR really, but can we throw together some info about the A100 minimum CUDA versions or a sensible minimum which may just be the 11.X on system at present.

jkwmoore (approved):

PR #1720, opened 2023-04-14
R docs for Stanage

desmfryan:
R docs


jkwmoore:
Can we add a warning about disparate node architecture?

R installs sometimes compiling software is going to cause installed packages to cease working when compiled on dissimilar CPUs to those it runs on.

PR #1736, opened 2023-04-19
Add redirect for software

Carldkennedy:


willfurnass:
Don't think there's any harm in just breaking the link now the pilot is over - might convey to pilot users that the docs have changed significantly recently.

PR #1742, opened 2023-04-20
Add info on installation of software with dependent modules.

jkwmoore:
This PR also corrects a spurious TOCtree inclusion and fixes a reference to our live clusters.

### DIFF
--- a/hpc/modules.rst
+++ b/hpc/modules.rst
@@ -340,8 +340,13 @@ Once the software is installed, each time you use the software you must first lo
 
 For more detailed information on the software installation process, please see: :ref:`installing-personal-software-installations`.
 
-If there are no preconfiguration steps available to generate a suitable **makefile** based on the current shell environment after loading modules or an example **makefile** for editing is provided, 
-you will have to construct/edit your own customised **makefile** which may have to reference specific libraries and paths.
+If:
+
+* There are no preconfiguration steps available to generate a suitable **makefile** based on the current shell environment after loading modules.
+* An example **makefile** for editing is provided.
+* No **makefile** is provided.
+
+You will have to construct/edit your own customised **makefile** which may have to reference specific libraries and paths.
 
 In this case, you can use the ``module show modulename`` command to show how the module file for your loaded software module/s are interacting with your shell environment to populate the ``$PATH``, ``$LD_LIBRARY_PATH`` 
 and other environment variables.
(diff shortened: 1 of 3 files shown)


hpc/modules.rst
```diff
@@ -317,6 +328,28 @@ If you no longer want to to have access to your own module files then you can ru
...
 
+-----
+
+Compiling software dependent on modules
+---------------------------------------
+
+In most cases, if you are compiling software with dependencies on modules the only actions you need to take are to load the required modules, run any ``./configure`` or **CMake** steps 
+and then run the ``make``, ``make check`` (if available) and ``make install`` commands to build, check and install the software.
+
+Once the software is installed, each time you use the software you must first load the modules used to compile it to make the required libraries and other files used during the compilation available to the program.
```
Carldkennedy, line 339:
```suggestion
Once the software is installed, each time you use the software you must first load the modules used to compile it. This is necessary to make the required libraries and other files used during the compilation available to the program.
```

hpc/modules.rst
```diff
@@ -317,6 +328,28 @@ If you no longer want to to have access to your own module files then you can ru
...
+
+For more detailed information on the software installation process, please see: :ref:`installing-personal-software-installations`.
+
+If there are no preconfiguration steps available to generate a suitable **makefile** based on the current shell environment after loading modules or an example **makefile** for editing is provided,
```
Carldkennedy, line 343:
```suggestion
If there are no preconfiguration steps available to generate a suitable **makefile** based on the current shell environment after loading modules, or if an example **makefile** for editing is not provided, 
```

Carldkennedy (commented):
LGTM with a  few small changes

hpc/modules.rst
jkwmoore: I did mean if a editable file is provided here. 

One not being provided would also require them to get the same additional info mind you, but it makes  for a really long sentence. Refactor into bullets?

jkwmoore: e.g.

If:


- There are no preconfiguration steps available to generate a suitable makefile based on the current shell environment after loading modules.

- An example makefile for editing is provided.

- No makefile is provided.

You will have to construct/edit your own customised makefile which may have to reference specific libraries and paths.

Carldkennedy: Looks good. Maybe

You will need to construct/edit your own customised makefile which may have to reference specific libraries and paths if:
.
.
.

Carldkennedy: Or the way you just did it :)

Carldkennedy (approved):
LGTM

PR #1746, opened 2023-04-20
Add SoX docs

Carldkennedy:


jkwmoore:
Given the use of the 'almost a standard template' do you think its worth adding a standard template to your other PR - https://github.com/rcgsheffield/sheffield_hpc/pull/1741

Carldkennedy:
Sounds like a great idea. Although a standard template in the future will contain tabs for all clusters @jkwmoore

PR #1752, opened 2023-04-24
Stanage: add libunistring compatible with GCC 10.2.0

willfurnass:
As we were missing a version compatible with one of our main supported toolchains.

### DIFF
--- a/stanage/software/libs/libunistring.rst
+++ b/stanage/software/libs/libunistring.rst
@@ -4,7 +4,7 @@ libunistring
 ============
 
 .. sidebar:: libunistring
-    
+
     :Versions: 0.9.10, 1.0
     :Documentation:  http://www.gnu.org/software/libunistring/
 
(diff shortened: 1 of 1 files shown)


stanage/software/libs/libunistring.rst
```diff
@@ -22,13 +22,14 @@ manipulating C strings according to the Unicode standard.
...
 Usage
 -----
-To make the library available, run one of the following: 
+To make the library available, run one of the following:
+
+.. code-block::
 
-.. code-block:: 
-         
       module load libunistring/0.9.10-foss-2019b
```
jkwmoore, line 29: Did we want to remove this 2019b version? I know we're wanting to shift people to the more modern versions but it is still on system as it stands.

willfurnass: Looks like it was added to satisfy:

- libunistring/0.9.10-GCCcore-7.3.0
    - Guile/1.8.8-GCCcore-7.3.0
        - libmatheval/1.1.11-GCCcore-7.3.0.

@nikishe Know if/why libmatheval needed?

nikishe: No sorry will , I would say leave it on....???

PR #1762, opened 2023-04-27
Amend Stanage Fluent SMP job script.

jkwmoore:
It appears (but I have been unable to confirm) that some models may be struggling to spawn using the same options as the MPI job as a result of asymmetric memory needs from the Fluent host process (which is likely as a result of multiple tasks and ring fenced memory per core.)

Leaving this in draft until I can be certain of the behaviour I am seeing but have currently been unable to reproduce the issue one of our users is having which was corrected by swapping to the single task with multiple cores SMP approach.


willfurnass:
I wonder if anyone's looked into the use of Slurm heterogeneous jobs [1] for Fluent

[1] https://slurm.schedmd.com/heterogeneous_jobs.html

jkwmoore:
I have, but not in any detail (yet). It **should** offer a preferred solution for optimized memory requests in the MPI jobs where the host process does want more memory than its children.

jkwmoore:
The amendment here certainly seems to be the most optimal although I've still been unable to replicate the issues reported by the user with the previous commands. 

Would like to merge sooner rather than later.

PR #1772, opened 2023-05-02
Fix broken diagram display on filestores page.

jkwmoore:
It appears that the javascript being used for the tabbing is also breaking the javascript for the Diagrams.net flow diagram.

tl;dr - we bypass the problem by replacing the raw HTML javascript XML diagram combo with a straight SVG.

The raw HTML import preserves the ability to click the links within the SVG rather than opening the SVG itself (if inlined as a Sphinx RST image.)


jkwmoore:
image

Now looks the exact same ^ as before.

nikishe (approved):

PR #1773, opened 2023-05-03
Linking NCDU stannage to the FAQ page

nikishe:
This eventually will be one link , but for now the redundancy is a necessary evil.


jkwmoore (approved):
Made a couple of quick corrections, LGTM.

PR #1775, opened 2023-05-03
Update OpenFOAM Stanage

Carldkennedy:


willfurnass (approved):

stanage/software/apps/openfoam.rst
```diff
@@ -3,10 +3,10 @@ OpenFOAM
 
 .. sidebar:: OpenFOAM
 
-   :Versions: 8.0, v2012
+   :Versions: 8.0, v2012, v2206
    :URL: https://openfoam.org/ or https://www.openfoam.com/
    :Documentation: https://cfd.direct/openfoam/user-guide or https://www.openfoam.com/documentation/overview
-   :Dependencies: Easybuild foss/2020a toolchain, NCurses 6.2, METIS 5.1.0, SCOTCH 6.0.9, CGAL 4.14.3 and Paraview 5.8.0
+   :Dependencies: Easybuild foss/2022a toolchain see :ref:`stanage_eb_toolchains`
```
jkwmoore, line 9: Foss 2022a, 2020a and 2020b here, should we just call it Easybuild foss?

Not for this PR but I also think our development toolchains page needs a bit of work.

jkwmoore (changes requested):
Pending changes on line 9 and the tests on system confirming it work being complete. LGTM.

stanage/software/apps/openfoam.rst
```diff
@@ -6,7 +6,7 @@ OpenFOAM
    :Versions: 8.0, v2012, v2206
    :URL: https://openfoam.org/ or https://www.openfoam.com/
    :Documentation: https://cfd.direct/openfoam/user-guide or https://www.openfoam.com/documentation/overview
-   :Dependencies: Easybuild foss/2022a toolchain see :ref:`stanage_eb_toolchains`
+   :Dependencies: Easybuild foss/2020a, 2020b and 2022a toolchains see :ref:`stanage_eb_toolchains`
```
jkwmoore, line 9: I was more getting at: 
```suggestion
   :Dependencies: Easybuild foss toolchain see :ref:`stanage_eb_toolchains`
```

stanage/software/apps/openfoam.rst
willfurnass: Think I agree with @jkwmoore 's suggestion here.

stanage/software/apps/openfoam.rst
willfurnass: Do we still want to mention NCurses, METIS, SCOTCH, CGAL and Paraview if they're all run-time dependencies?

jkwmoore: Some of these (METIS, SCOTCH, CGAL) will be depending on what code you invoke so it very well might be worth outright telegraphing that people can check them with a module load or inspecting the Easyconfig.

iirc Paraview is just wrapped in Parafoam with some workflow bits to help out but I might be wrong. Not sure on NCurses, but I expect it is also runtime depd.

jkwmoore: In a general sense I think we might need to do more to assist users in figuring out dependencies for Easybuild/modules software and I put some effort into the modules page not so long ago around that.

We may want to generate a short importable section with the crib notes of how to check this for dependency heavy software / modules like this.

Carldkennedy:
All examples in docs completed successfully

jkwmoore (approved):

PR #1778, opened 2023-05-04
H100 nodes: specs and slurm config

willfurnass:
- [x] Hardware specs
- [x] Slurm config


willfurnass:
Not to be merged until just before H100 nodes made available to end users.

Carldkennedy (approved):
LGTM

nikishe (approved):
LGTM

stanage/cluster_specs.rst
```diff
@@ -129,14 +156,14 @@ Total capacity
...
 
   * Intel Cores: 11264.
-  * AMD Cores: 768.
+  * AMD Cores: 1344.
   
-* Total memory:  83968 GiB.
-* GPUs: 64.
+* Total memory:  90112 GiB.
+* GPUs: 76.
 * Fast network filesystem (`Lustre`_):  2 PiB.
```
jkwmoore, line 167: Can we add an invisible RST comment here on the calculation method we're using for nodes, cores, memory etc... e.g. with / without GPU nodes, with/without "private nodes" etc...

willfurnass: Good idea; can handle via separate Issue/PR.

willfurnass: https://github.com/rcgsheffield/sheffield_hpc/issues/1795

jkwmoore (approved):

willfurnass:
- [ ] Add info on NVlink vs PCIe: H100s might not always be faster
       - NVLink + 2x A100 gives 3-4x inter-GPU bandwidth vs PCIe + 2x H100
       - but inter-GPU latencies actually lower for 2x H100

willfurnass:
Rebased on `master`

willfurnass:
@Carldkennedy and @nikishe: Please can you review the new content on limitations of H100 support in latest PyTorch and Tensorflow releases: 

```
referenceinfo/imports/software/tensorflow/compat-table-import.rst
stanage/software/apps/pytorch.rst
stanage/software/apps/tensorflow.rst
```

willfurnass:
Also, where should we stick comments about perf differences between H100s and A100s being greatest when probs are compute-bound (rather than mem-bound) and with larger prob sizes?

stanage/software/apps/pytorch.rst
```diff
@@ -48,17 +53,19 @@ Then PyTorch can be installed by the following ::
...
 
    # Activate the 'pytorch' environment
    source activate pytorch
 
-   # Install PyTorch
-   pip install torch torchvision
+   # Install the latest stable PyTorch release if you only want to run PyTorch using A100 GPUs
+   python -m pip install torch torchvision
+   # Or use a nightly PyTorch build instead if you want to be able to use H100 GPUs also / as well.
+   # python -m pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu118
```
Carldkennedy, line 68: `module load cuDNN/8.7.0.84-CUDA-11.8.0` for pytorch-nightly build?

Carldkennedy (commented):
LGTM other than one comment

Carldkennedy:
> Also, where should we stick comments about perf differences between H100s and A100s being greatest when probs are compute-bound (rather than mem-bound) and with larger prob sizes?

Maybe a note imported in GPU computing and below example batch scripts in  R, lammps, gromacs, pytorch and tensorflow?

jkwmoore:
I've amended slightly to fully specify the GPU GRES with a mind to us 

A: sticking to this approach in future to ensure user's decouple/select GPUs correctly.
B: can add a "Any GPU node(s)" option, when the H100s are moved into the main GPU partition.

With respect to Bessemer and the temporarily available  V100 16GB models, we can then be opinionated via `--gres=gpu:v100-16GB:1` or similar.

We may wish to align the release of the H100 with any amendment to the Bessemer setup in order to use the same notification email (but since any changes to Bessemer are a bit of a breaking change we need to give a week's notice!)

jkwmoore:
I have amended this to add explicit warnings about our "opt-in" plan of action.

Carldkennedy (approved):
LGTM

PR #1780, opened 2023-05-04
Stanage h100 gencodes

ptheywood:
Adds CUDA gencode docs for the upcoming H100 nodes (compute capability 90), and adds a few links.

Also fixes a minor RST issue in the ncu section.

Doesn't include any recommendations for using which CUDA version(s) for best H100 support (11.8+ minimum, 12.0+ for full features).

---

I'll double check this works / is correct when possible, prior to marking as ready for review.

### DIFF
--- a/stanage/software/libs/cuda.rst
+++ b/stanage/software/libs/cuda.rst
@@ -141,13 +141,26 @@ i.e. ``-gencode=arch=compute_80,code=compute_80``.
 
 The minimum specified virtual architecture must be less than or equal to the `Compute Capability <https://developer.nvidia.com/cuda-gpus>`_ of the GPU used to execute the code.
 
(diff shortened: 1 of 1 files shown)


ptheywood:
I've now tested this on the h100 nodes, and the suggested gencode arguments generate executable code with 12.0, 11.8, 11.7 and 10.2.

nikishe (approved):

ptheywood:
This is probably worth merging now that the H100s are avialable.

jkwmoore:
> This is probably worth merging now that the H100s are avialable.

Ah, this was meant to be merged on the Friday (with the other PRs) with the availability of the H100s on the following Monday.

Merging now.

PR #1787, opened 2023-05-09
Update MPI FAQ with export ALL

Carldkennedy:


FAQs.rst
```diff
@@ -704,6 +704,21 @@ that is compatible with the PMI2 or PMI-X library used by the Slurm job schedule
...
+On Bessemer and Stanage in batch scripts you should use the ``--export=ALL`` option with the ``srun`` command, 
+which tells SLURM to export all of the current shell environment variables to the job environment.
+
+.. code-block::
+
+        srun --export=ALL my_program
+
+This is important because many applications and libraries rely on environment variables to locate their dependencies, such as shared libraries.
+
+
```
jkwmoore, line 716: Should we give an example here of what happens as a result of forgetting to use the export all option?

Carldkennedy: Agreed that's a good idea @jkwmoore

Carldkennedy: Ah just realised I used VASP as an example. Since a licence is needed to use this module would it be better to use another example? Just in case they want to run through it themselves @jkwmoore

jkwmoore: Probably not that important since people are unlikely to intentionally replicate something designed to fail. If we want to be thorough though we can probably swap this out for a simple mpirun helloworld example.

Carldkennedy: Replaced example with OpenMPI hello world @jkwmoore

FAQs.rst
```diff
@@ -704,6 +704,39 @@ that is compatible with the PMI2 or PMI-X library used by the Slurm job schedule
 This is the case for the administrator-provided versions of OpenMPI and Intel MPI on Bessemer and Stanage;
 again, no extra configuration is required by the end user.
 
+On Bessemer and Stanage in batch scripts you should use the ``--export=ALL`` option with the ``srun`` command, 
+which tells SLURM to export all of the current shell environment variables to the job environment.
```
jkwmoore, line 708:
```suggestion
which tells Slurm to export all of the current shell environment variables to the job environment.
```

FAQs.rst
```diff
@@ -704,6 +704,39 @@ that is compatible with the PMI2 or PMI-X library used by the Slurm job schedule
...
+.. code-block:: console
+
+        srun --export=ALL my_program
+
+This is important because many applications and libraries rely on environment variables to locate their dependencies, such as shared libraries.
+
+Take, for instance, if we were to submit this :ref:`OpenMPI non-interactive hello world job <batch_openmpi_stanage>` without the ``--export=ALL`` option, i.e:
+
+.. code-block:: console
+    :emphasize-lines: 5
```
jkwmoore, line 719:
```suggestion
       :emphasize-lines: 5
```

jkwmoore, line 719:
```suggestion
       :emphasize-lines: 5
```

FAQs.rst
```diff
@@ -704,6 +704,39 @@ that is compatible with the PMI2 or PMI-X library used by the Slurm job schedule
...
+        
+        #!/bin/bash
+        #SBATCH --nodes=1
+        #SBATCH --ntasks-per-node=8
+        module load OpenMPI/4.1.4-GCC-12.2.0
+        srun hello
+
+We would encounter an error message containing:
```
jkwmoore, line 727:
```suggestion
On the Stanage cluster, we would encounter an error message containing:
```

FAQs.rst
```diff
@@ -704,6 +704,39 @@ that is compatible with the PMI2 or PMI-X library used by the Slurm job schedule
...
+
+.. code-block:: console
+
+        [node140.pri.stanage.alces.network:12429] PMIX ERROR: NOT-FOUND in file client/pmix_client.c at line 562
+
+While loading the OpenMPI module will set the variable ``SLURM_MPI_TYPE=pmix_v4``, 
+when srun is initiated it creates a new environment. Since we haven't instructed it to export the environment variables to this new environment,
```
jkwmoore, line 734:
```suggestion
when ``srun`` is initiated it creates a new environment. Since we haven't instructed it to export the environment variables to this new environment,
```

stanage/software/apps/vasp.rst
```diff
@@ -39,6 +39,8 @@ VASP Pseudopotentials can be found in the directory pointed to by the ``$PSEUDOP
 
 Note that the ``$PSEUDOPOTENTIAL_DIR`` environment variable is only defined *after* you have loaded a VASP environment module.
 
+.. _vasp_batch_stanage:
```
jkwmoore, line 42: Is this needed anymore?


```suggestion
```

Carldkennedy: No it's not. Thanks @jkwmoore

willfurnass (approved):

nikishe (approved):
LGTM (liked the example)

PR #1790, opened 2023-05-10
Promote use of ~ and/or $HOME

nikishe:
I feel This might be too big for a note but couldnt figure how else to place it

### DIFF
--- a/hpc/filestore.rst
+++ b/hpc/filestore.rst
@@ -118,7 +118,7 @@ All users have a home directory on each system:
   | Bessemer & ShARC  |``/home/$USER``         |
   +-------------------+------------------------+ 
 
-  To be cluster agnostic across all three we recommend you use ``~`` and ``$HOME``. This will resolve to the right path no matter the cluster. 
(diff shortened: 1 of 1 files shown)


hpc/filestore.rst
```diff
@@ -106,6 +106,52 @@ All users have a home directory on each system:
...
+
+  +-------------------+------------------------+
+  | Cluster           | Path                   |
+  +===================+========================+
+  | Stanage           |``/users/$USER``        |
+  +-------------------+------------------------+
+  | Bessemer & ShARC  |``/home/$USER``         |
+  +-------------------+------------------------+ 
+
+  To be cluster agnostic across all three we recommend you use ``~`` and ``$HOME``. This will resolve to the right path no matter the cluster.
```
Carldkennedy, line 121:
```suggestion
  To ensure that your code is compatible with all three clusters, we suggest using the symbols "~" or "$HOME" to represent the home directory. This approach ensures that the correct path is used regardless of the cluster you are working on, making your code more portable and agnostic to the specific cluster environment.
```

Carldkennedy (approved):

PR #1794, opened 2023-05-11
Small correction toolchains on Stanage

Carldkennedy:


nikishe (commented):
LGTM

willfurnass:
`fosscuda` not installed at present? 

Suggest we only install more in the way of older CUDA tooling if we really need to.  Folks will get best perf if they adopt CUDA >= 11 to make best use of the A100s and >= 11.8 to make best use of the H100s.

Carldkennedy:
> `fosscuda` not installed at present?
> 
> Suggest we only install more in the way of older CUDA tooling if we really need to. Folks will get best perf if they adopt CUDA >= 11 to make best use of the A100s and >= 11.8 to make best use of the H100s.

You're right @willfurnass, not installed. I'll remove those sections.

stanage/software/development/toolchains.rst
```diff
@@ -8,7 +8,7 @@ was installed using `EasyBuild <https://easybuild.readthedocs.io/>`__,
 which is a tool for building and installing software on HPC systems.
 
 EasyBuild groups core sets of dependencies as what it calls *toolchains*.
-The three supported toolchains on Stanage are:
+The two supported toolchains on Stanage are:
```
jkwmoore, line 11: Should this really be "The two main toolchains 'with support|available' on Stanage are" ? Is this support from us or just availability on the cluster?

If we mean availability we should say that instead I think.

jkwmoore (changes requested):
LGTM - we may just want to decide on whether the suggested amendment is needed.

nikishe (approved):
LGTM (Post jame's suggestion)

stanage/software/development/toolchains.rst
Carldkennedy: The two main toolchains available with support on Stanage are: ?

jkwmoore: Well I suppose it depends mainly on whether you fellas want to explicitly say you will support those toolchains, if yes, how? The latest 2 years worth etc...

I'm not averse to saying their available but at some point we want to explicitly declare our intentions in terms of EB toolchain support formally.

This page seems like the sensible place?

willfurnass: Suggest getting this merged and revisiting this conversation via separate Issue/PR.

PR #1821, opened 2023-06-07
Add Stanage host keys to FAQ.

jkwmoore:
See above.


jkwmoore:
At some point we probably want to amend these host key blocks as an import for each cluster. But I will leave that up to the next maintainer who is going to refactor the main SSH connection info sections to include manually checking the keys and including the host keys alongside.

willfurnass (approved):

Carldkennedy (approved):

nikishe (approved):

PR #1839, opened 2023-07-24
Introduce 'seff' script to Slurm clusters

jkwmoore:


jkwmoore:
I was not descriptive with commit messages here so please squash merge!

jkwmoore:
n.b. if approved merging at 11am on Friday is ideal - but the comms email has already been sent so people should be aware that the commands won't be available until that time.

willfurnass (approved):

PR #1842, opened 2023-07-26
1796 matlab documentation on stanage

nikishe:


stanage/software/apps/matlab.rst
```diff
@@ -0,0 +1,136 @@
...
+
+.. _matlab_stanage:
+
+|softwarename|
+==========================================================================================================
+
+.. sidebar:: MATLAB
+
+   :Versions:  2022a
+   :Support Level: FULL
```
jkwmoore, line 12: From the user perspective, what does this mean? Do we have a blurb somewhere on our agreed policy and what the support levels are?

stanage/software/apps/matlab.rst
```diff
@@ -0,0 +1,136 @@
...
+-----------------
+After connecting to Stanage,  start an interactive session with the ``srun --pty bash –i`` command.
+
+The latest version of MATLAB (currently 2022a) is made available by running:
+
+.. code-block:: bash
+
+   module load MATLAB/2022a
+
+You can then run MATLAB by entering ``matlab &``.
```
jkwmoore, line 30: Does this normally spawn a GUI? If so this won't spawn a GUI with the current Stanage setup.

stanage/software/apps/matlab.rst
```diff
@@ -0,0 +1,136 @@
...
+.. code-block:: bash
+
+   sbatch parallel_example.slurm
+
+
+The MATLAB script, ``parallel_example.m``, 
+creates 200 square (500 x 500) matrices comprised of random values,
+calculates the eigenvalues of each 
+and records the maximum eigenvalue for each matrix in the array ``max_eigenvals``.
+
```
jkwmoore, line 126: As this section is detailing SMP style parallelization - do we want to mention a lack of MPI support at this time?

Longer term do we want to include a section on MPI supported parallelization (iirc on Des's comments probably no!)

willfurnass: Suggest filing separate Issue for that point and just getting this merged.

stanage/software/apps/matlab.rst
willfurnass: Suggest tacking as separate Issue.

stanage/software/apps/matlab.rst
willfurnass: Yes, we should note that this starts a MATLAB command-line REPL on Stanage.  

Also, I suggest ditching the `&` for now.

stanage/software/apps/matlab.rst
nikishe: Ive taken it ou , as it applies to us rather than users

Issue #1857, opened 2023-08-07
Stanage: default mem per CPU is 4016 MB, not 2048 MB


willfurnass:
Need to update https://docs.hpc.shef.ac.uk/en/latest/hpc/Choosing-appropriate-resources.html#memory-allocation-limits and possibly other pages too.

PR #1859, opened 2023-08-08
Stanage default memory correction.

jkwmoore:
Also includes a minor clarification on Bessemer given Stanage can do SMP and MPI vs Bessemer's SMP only.

### DIFF
--- a/referenceinfo/imports/scheduler/memory_allocation_limits_table_import.rst
+++ b/referenceinfo/imports/scheduler/memory_allocation_limits_table_import.rst
@@ -15,9 +15,9 @@
      - 251 GB
      - 1007 GB 
      - 2014 GB
-     - 2 GB / 251 GB
-     - 2 GB / 251 GB (SMP) ~74404 GB (MPI)
-     - **Per job basis** ``--mem=<nn>``
+     - 4 GB / 251 GB
+     - 4 GB / 251 GB (SMP) ~74404 GB (MPI)
+     - **Per node basis** ``--mem=<nn>``
 
    * - SLURM (Bessemer)
      - 192 GB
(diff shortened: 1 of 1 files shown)


jkwmoore:
Closes https://github.com/rcgsheffield/sheffield_hpc/issues/1857

jkwmoore:
n.b. as this is corrected in the import it handles it everywhere and I have checked the job submission pages for further errors and not seen them.

referenceinfo/imports/scheduler/memory_allocation_limits_table_import.rst
```diff
@@ -15,17 +15,17 @@
      - 251 GB
      - 1007 GB 
      - 2014 GB
-     - 2 GB / 251 GB
-     - 2 GB / 251 GB (SMP) ~74404 GB (MPI)
-     - **Per job basis** ``--mem=<nn>``
+     - 4 GB / 251 GB
```
willfurnass, line 18: 4096 MB != 4016 - could catch users out.

jkwmoore: True, but unlikely. As this is the default value my expectation is that people are largely not going to check it and set their jobs to use  4096.

That said, I prefer the technical correctness. Prefer breaking the table convention and using Mb or asking Alces to set it to 4096? (I'm more inclined to the latter.)

jkwmoore: Given chat irl - given memory allocation at 64 cores on nodes due to Alces memory reservation on hosts with 4096MB will never run (not enough memory) erring on the side of breaking table convention.

referenceinfo/imports/scheduler/memory_allocation_limits_table_import.rst
```diff
@@ -15,17 +15,17 @@
...
+     - 4 GB / 251 GB (SMP) ~74404 GB (MPI)
```
jkwmoore, line 19:
```suggestion
     - 4016 MB / 251 GB
     - 4016 MB / 251 GB (SMP) ~74404 GB (MPI)
```

PR #1861, opened 2023-08-10
Add GA tracking ID

Carldkennedy:
Example of adding a Google Analytics Tracking ID to the site

It's recommended to use a shared google account to produce the Tracking ID


jkwmoore:
We may want to use this plugin in case the theme ever changes:
https://pypi.org/project/sphinxcontrib-googleanalytics/

Carldkennedy:
@jkwmoore Implemented sphinxcontib extension as per your suggestion.

Tracking ID is inserted in the headers of all html pages 

`grep -r "id=G-HTK1JMDS73" sheffield_hpc/_build/html/ | wc -l`
354

themes/sheffieldhpc/layout.html
```diff
@@ -42,4 +42,5 @@
    {%- endif %}
   </div>
 </footer>
+
```
Carldkennedy, line 45:
```suggestion
```

nikishe (approved):
LGTM

PR #1868, opened 2023-08-24
Add basic info for citing and for letters of support

willfurnass:
@jonbarker68 Any comments on this?


willfurnass:
NB we do want better text for citing the HPC systems but can revisit that after this PR is merged.

Carldkennedy (approved):
LGTM

PR #1886, opened 2023-10-09
Memory call

nikishe:


jkwmoore:
Why 16GB? For a given  think we've generally recommended taking at least as much RAM as the GPU has VRAM?

nikishe:
> Why 16GB? For a given think we've generally recommended taking at least as much RAM as the GPU has VRAM?

I have edited it to 82gb

Carldkennedy (approved):
LGTM

PR #1889, opened 2023-10-18
MobaXterm terminal: warn that causes issues w/ Stanage + MFA

willfurnass:
As discovered by a user v. recently.

Connecting via profiles and SSHing from a Powershell prompt both work fine though.


jkwmoore (approved):

jkwmoore:
Before merging this, do we also want to include a warning in the specific Stanage FAQ / quirks section?

PR #1910, opened 2023-12-01
Refactor ShARC documentation into new "decommissioned" section

Carldkennedy:
Move ShARC to new decommissioned section 
Add stubs for relocated ShARC docs
Move Iceberg into decommissioned section 
Add watermark to decommissoned and sharc specific message, removed globally.


willfurnass:
What. A. PR.

willfurnass:
Links to a bunch of R install scripts and modulefiles now broken.

nikishe:
Tabs still intact but this might be good for the ones transitioning. Thought if we decide to delete , do we have 2 copies , one in decomisioned with just "sharc" tab and update live with the other 2 / Do we just delete the tab (Might be a pain for users but atleast admins can always pull from git v0.0.1 if we want to reference ?

image

nikishe:
should this also be moved to decomissioned or  have a new addition

image

Ideas for the new addition 
image

nikishe:
SGE links in "Job Submission and Control" are broken , even though pointing to correct decommisioned section

**decommissioned/sharc/sge.html#sge-info**

Carldkennedy:
> should this also be moved to decomissioned or have a new addition

Will do in separate PR

jkwmoore:
Can we get the Global ShARC decom warning removed from all pages and only applied to the ShARC subsection of the "Decommissioned" tab?

nikishe (approved):

jkwmoore (changes requested):
LGTM. I'd just want the sections which apply the watermarking be done as an import so if you wanted to amend all decommissioned pages in future, you could edit the import once rather than having to add more random bits in / around:


```
.. raw:: html

   <link rel="stylesheet" href="/_static/css/watermark.css">
   ```

willfurnass:
@Carldkennedy Wanting any more reviews of this?  If so, are you able to summarise the changes inc intent?  Is the first comment in this PR still comprehensive/accurate?

Carldkennedy:
> LGTM. I'd just want the sections which apply the watermarking be done as an import so if you wanted to amend all decommissioned pages in future, you could edit the import once rather than having to add more random bits in / around:
> 
> ```
> .. raw:: html
> 
>    <link rel="stylesheet" href="/_static/css/watermark.css">
> ```

@jkwmoore As a separate import or add to the existing import? Maybe cleaner to have in a single import?

jkwmoore:
Given the current imports only apply to the ShARC section, this would need to be in a new one.

Ideally the import for CSS would also be as close to the top as possible.

jkwmoore:
> > @Carldkennedy Wanting any more reviews of this? If so, are you able to summarise the changes inc intent? Is the first comment in this PR still comprehensive/accurate?
> 
> Squashed a few commits more accurately. Should be ready to merge if happy

LGTM, yolo.

PR #1913, opened 2023-12-01
Example watermarking 

jkwmoore:
Just an example for Carl to look at.

See also:

image
image


Carldkennedy:
Thanks @jkwmoore. Added to https://github.com/rcgsheffield/sheffield_hpc/pull/1910

PR #1957, opened 2024-01-16
Add resetenv path

Carldkennedy:
Will reference full path to resetenv on both clusters. 

 - [ ] refactor resetenv to it's own path

### DIFF
--- a/FAQs.rst
+++ b/FAQs.rst
@@ -33,6 +33,7 @@ If you hit this problem you can run the command ``resetenv`` which will restore 
                         .. code-block:: console
 
                                 . /usr/local/scripts/resetenv
+
 ------
 
 I can no longer log in
--- a/FAQs.rst
+++ b/FAQs.rst
@@ -25,14 +25,14 @@ If you hit this problem you can run the command ``resetenv`` which will restore 
 
                         .. code-block:: console
 
-                                . /opt/site/etc/profile.d/060-bash-shell-helpers.sh && resetenv
+                                /opt/site/bin/resetenv
 
 
                 .. group-tab:: Bessemer
 
                         .. code-block:: console
 
-                                . /usr/local/scripts/resetenv
+                                /usr/local/scripts/resetenv
 
 ------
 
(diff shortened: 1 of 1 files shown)


FAQs.rst
```diff
@@ -15,6 +15,25 @@ Symptoms include many of the commands not working and just ``bash-4.1$`` or ``sh
 This may be because you've deleted your ``.bashrc`` and ``.bash_profile`` files - these are 'hidden' files which live in your home directory and are used to correctly set up your shell environment.
 If you hit this problem you can run the command ``resetenv`` which will restore the default files, then you should logout and log back in.
 
+.. note:: 
+
+        Alternatively, you can run the resetenv command directly:
```
jkwmoore, line 20: This isn't an alternative per se - this is what they need to do if the command does not work when invoked with ``resetenv`` due to truly bodging their $PATH.

Can we amend to have this as a further step if ``resetenv`` does not work.

willfurnass:
I wonder we'd want to take a simpler approach of 
1. always ensuring `resetenv` is a standalone executable script
2. always referencing it by absolute path in our docs

Thoughts?

jkwmoore:
1. No objections, current implementation is as a function for Stanage to keep everything in one file, but refactoring it to its own file is trivial.

2. So long as we make it clear things are different per cluster.

Carldkennedy:
Are we going with standalone executable script on both clusters (addressing resetenv with full path)?
Could this be the same path on both clusters?

willfurnass:
My preference would be for a very simple stand-alone script per system, but ensuring the same path across both systems might be tricky given we don't have privileged access to Stanage to add things to e.g. `/usr/local/bin/`.

@jkwmoore is to create a separate PR in a mgmt repo to cover ^ for Stanage.

jkwmoore:
Pending a review on a PR we have for Stanage to make the changes needed.

`resetenv` will be found in `/opt/site/bin/resetenv` for Stanage.

jkwmoore:
This is now available on Stanage at:

```bash
[user@login1 [stanage] ~]$ which resetenv
/opt/site/bin/resetenv
```

jkwmoore (approved):

PR #1961, opened 2024-01-23
Add STAR to Stanage

Carldkennedy:


stanage/software/apps/STAR.rst
```diff
@@ -0,0 +1,138 @@
...
+  gunzip GRCh38.primary_assembly.genome.chr19.fa.gz
+  
+  # download the annotations that correspond to it 
+  wget https://ndownloader.figshare.com/files/14658830 && \
+  mv 14658830 gencode.v29.primary_assembly.annotation.chr19.gtf.gz && \
+  gunzip gencode.v29.primary_assembly.annotation.chr19.gtf.gz
+
+
+The following is an example batch submission script, ``my_job.sh``, to run the executable ``STAR``.
+The script requests 4 cores using the OpenMP parallel environment with a runtime of 30 minutes and
```
willfurnass, line 66: Think 'OpenMP parallel environment' has been carried over from the ShARC/SGE docs.

stanage/software/apps/STAR.rst
```diff
@@ -0,0 +1,138 @@
...
+
+--------
+
+Interactive usage
+-----------------
+
+After connecting to Stanage (see :ref:`ssh`),  start an interactive session with the 
+:code:`srun --pty bash -i` command.
+
+The latest version of STAR (currently version 2.7.10b) is made available with the command:
```
willfurnass, line 29: Latest on Stanage or latest from the STAR team?   Could just say `load STAR by running:`?

stanage/software/apps/STAR.rst
willfurnass: STAR dynamically links against `libgomp` (OpenMP lib) and does multithreading

```
[someone@bessemer-node002 ~]$ ldd $(which STAR)
...
        libgomp.so.1 => /usr/local/packages/live/eb/GCCcore/9.3.0/lib64/libgomp.so.1 (0x00002b70670f8000)
...
```
so I think it makes sense to request multiple cores for a single task rather than multiple tasks.

willfurnass (changes requested):
See comment re requesting cores per task rather than multiple tasks.

stanage/software/apps/STAR.rst
willfurnass: Bessemer page could also do with being updated in these regards.

stanage/software/apps/STAR.rst
jkwmoore: I think historically, (at least me) tried to ensure that users were rail roaded into selecting specific versions to avoid them getting used to no versions which could result in unexpected changes in the future with newer modules.

Carldkennedy:
Thank you @willfurnass. s
Suggested changes implemented.

PR #1962, opened 2024-01-24
Add flight port forwarding password note

Carldkennedy:
Clearly indicate which password is needed at each step


jkwmoore:
Added a new commit to really try to keep users on the rails and enter the right stuff. Let me know what you think / feel free to edit or disagree with my changes.

nikishe (approved):

Carldkennedy:
Looks good. Thanks @jkwmoore

PR #1964, opened 2024-01-25
Increase opacity of watermark

Carldkennedy:
Slightly increases opacity of watermark


nikishe (approved):
LGTM

jkwmoore:
Just as a note for any future tweaking. 

The perception of the opacity here is greatly affected by monitor settings so we probably need to avoid constantly tweaking and re-tweaking when some monitors make it look bad.

PR #1966, opened 2024-01-25
1902 Dropdown rsync behaviour with trailing slashes

Carldkennedy:
Add ~sphinx_panels~ sphinx_design for dropdown capability - will make docs much easier to skim - more suitable for differing levels of experience
Add trailing slashes behaviour of rsync

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -27,6 +27,7 @@ extensions = [
     'sphinx_rtd_theme',
     'sphinx_copybutton',
     'sphinx_tabs.tabs',
+    'sphinx_panels',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/conf.py
+++ b/conf.py
@@ -27,7 +27,7 @@ extensions = [
     'sphinx_rtd_theme',
     'sphinx_copybutton',
     'sphinx_tabs.tabs',
-    'sphinx_panels',
+    'sphinx_design',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/hpc/transferring-files.rst
+++ b/hpc/transferring-files.rst
@@ -153,44 +153,45 @@ backup folders.
   
 .. _rsync_behaviour:
 
-.. dropdown:: :underline-bold:`rsync Behaviour with Trailing Slashes`
+.. note:: 
 
       Be cautious when specifying paths with or without trailing slashes. 
       Ensure that you understand how ``rsync`` interprets these slashes to prevent unintended outcomes. 
-
-
-      **With Trailing Slash on Source Directory**:
-
-      .. code-block::
-
-          rsync -av /source/directory/ /destination/directory
-
-      - When you use a trailing slash on the source directory it tells ``rsync`` to copy the **contents** of the source directory into the destination directory.
-
-      **Without Trailing Slash on Source Directory**:
-
-      .. code-block::
-
-              rsync -av /source/directory /destination/directory
-
-      - When you don't use a trailing slash on the source directory it tells ``rsync`` to copy the **source directory itself** and its contents into the destination directory.
-
-      **Trailing Slash on Destination Directory**:
-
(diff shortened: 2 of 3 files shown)


Carldkennedy:
Looks like we can't use sphinx panels as requires sphinx-rtd-theme <5 others need >5

jkwmoore:
As dropdowns are stuck at the moment do you want to bifurcate this PR and merge on the rsync bits?

Carldkennedy:
> As dropdowns are stuck at the moment do you want to bifurcate this PR and merge on the rsync bits?
@jkwmoore 
Replace sphinx-panels with sphinx-design. 
From their docs:
 "This package arose as an iteration on sphinx-panels, with the intention to make it more flexible, easier to use, and minimise CSS clashes with sphinx themes"

willfurnass:
A general comment: the more sphinx extensions we add into the mix the more likely we are to encounter issues in future when trying/needing to upgrade our sphinx ecosystem.

nikishe (approved):
LGTM as long as going forward we have reviews on the extensions we add on sphinx, as at the moment I believe we are sitting on 2 .

jkwmoore:
For the record I disagree with the use of drop-downs in the docs for the most part as I really do not see valid use cases. I only see:

1. More tech debt from a plugin support perspective.
2. Hidden info from users looking for info from search results, them now needing to unfurl everything to find info.
3. I have no idea what behaviour this plugin will do when the docs is exported to alternative formats like PDF. (Was this checked?)

The way I see things is:  the info is either important enough to be inline on the same page, or it isn't and should have a reference via hyperlink/warning/note or otherwise.

Carldkennedy:
- The sphinx-design repo seems popular and is fairly active.
- Sphinx search is not great so not sure if it's not finding the information or because the information is furled.
- Dropdowns should be reserved for non essential information.
In this example the main information is above the dropdown: "Be cautious when specifying paths with or without trailing slashes. Ensure that you understand how rsync interprets these slashes to prevent unintended outcomes."
The dropdown gives examples which a user could go elsewhere to find, but nice to not have to.
- Printing the webpage to pdf with the dropdown unfurled doesn't print the dropdown info (but here, the main information is.)
- In terms of searching within the page using 'ctrl+f' does unfurl dropdowns.

Agree that we could link elsewhere, for example "reference info section". This requires leaving the page, unless we use an extension like [hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/index.html) to show a floating window with the information. But again this won't print with the webpage.

I'm not sure how many users really would want to print our docs. And if only non-essential information is hidden I'm not sure it would be an hindrance. Similarly tabs will only print the currently active tab.

I think the main thing we're aiming for here is to have the information available but without cluttering the page. Skimmable.

jkwmoore:
> I think the main thing we're aiming for here is to have the information available but without cluttering the page. Skimmable.

Which I would argue should still be a separate referenced page rather than a drop-down. I personally just do not see dropdowns adding any valuable functionality in totality. 

If the info is not essential it shouldn't be on the page. I value the fact that any additional non-essential info is shifted to a separate page even if it adds an off page click rather than being furled/unfurled.

In this specific case, the doubling of notice boxes also looks untidy/cluttered too and I expect distracts the call to action.

 I would prefer a single caution box highlighting the dry run and risks with slashes with a single link highlighted to direct the user to a separate reference page, rather than the warning box, the note box and an furl/unfurl box.

> Similarly tabs will only print the currently active tab.

This is pretty awful to be honest. Besides the pain of refactoring, I really do like / prefer the idea of the branched per cluster (potentially colour coded background) subdirectories over any inline tabbing.

e.g.

https://docs.hpc.shef.ac.uk/en/latest/common
https://docs.hpc.shef.ac.uk/en/latest/stanage
https://docs.hpc.shef.ac.uk/en/latest/bessemer

PR #1981, opened 2024-02-05
Add recommended version callout

Carldkennedy:
Adds attention callout with information about peculiar behaviour of latest Anaconda version
Recommending 22.05


nikishe (approved):
LGTM

PR #1982, opened 2024-02-06
Update FAQs.rst - correct spelling error.

jkwmoore:


jkwmoore:
image

If the actions pass (which they will) I will merge.

PR #1987, opened 2024-02-09
Add custom google search readme info.

jkwmoore:
A bit more info on how this works rather than making you dig into the PR.


willfurnass:
Ta.  Any way of controlling frequency of spidering?  If so, worth noting in the README.  Couldn't spot anything in the CSE settings.

jkwmoore:
Sadly I don't see anything in the CSE stuff, but there may be something in the search console stuff.

willfurnass (approved):

PR #1990, opened 2024-02-13
Public permissions warning

fredsonnenwald:
Add a little note to the docs about how setting up a public shared area on par scratch really is public.
Some tidying of formatting.


jkwmoore (approved):
LGTM, but since this is self review:

image

Can someone else have a look to check we're still rendering everything.

nikishe (approved):
LGTM

PR #1992, opened 2024-02-15
Add inline warning about the need for interactive sessions.

jkwmoore:


Carldkennedy:
Missing the following line-  The following is example output:

Otherwise, LGTM

jkwmoore:
Nice catch, re-added.

PR #2007, opened 2024-03-04
Add salloc warning for changed Bessemer behaviour.

jkwmoore:
As part of C2402 477 - Change scheduler salloc command default behaviour.

We have changed the default salloc behaviour on Bessemer to now not only request an allocation of resources (as it does now), but to also then start a terminal shell within the **allocated** node.


jkwmoore (commented):
Added reviewers primarily to ensure that you're aware that this change has been made.

Going to merge immediately since this is already implemented.

nikishe (commented):
LGTM

PR #2000, opened 2024-02-29
First draft for port forwarding and ssh keys

nikishe:

PR #2011, opened 2024-03-06
First draft for port forwarding and ssh keys

jkwmoore:
Continuing from https://github.com/rcgsheffield/sheffield_hpc/pull/2000

@Carldkennedy @f-allian  probably worth re-adding any requested changes here.


hpc/sshkeys.rst
```diff
@@ -0,0 +1,97 @@
...
+
+    .. tab:: MobaXTerm
+
+        For users on windows using mobaxterm you can simply add your private key to any of the profiles you have set up as shown in the image below. This will open a file explorer window that will allow you to select your private key.
+
+        .. image:: /images/mobaxterm-PrivateSSHKey-setup.png
+
+    .. tab:: SSH-Agent
+
+        Check that you have ssh agent installed and running with the following command.
```
fredsonnenwald, line 61: What's the difference between SSH Agent, SSH-AGENT, and ssh agent?

stanage/software/apps/jupyter.rst
```diff
@@ -0,0 +1,131 @@
...
+   :URL: https://posit.co/products/open-source/rstudio/
+   :Documentation: https://jupyter-notebook.readthedocs.io/en/5.6.0/index.html
+
+
+Jupyter Notebook is an open source web application that you can use to create and share documents that contain live code, equations, visualisations, and text. 
+
+Installing and Running Jupyter Notebooks
+-----------------------------------------
+
+The HPC clusters are only able to host a Jupyter notebook work space in an interactive session. To install and run Jupyter please follow these steps.
```
fredsonnenwald, line 19: Will mentioned the possibility of just rendering a notebook?

stanage/software/apps/jupyter.rst
```diff
@@ -0,0 +1,131 @@
...
+.. note:: 
+
+    Please take note of the two addresses in the output. You will need them later. In this example it will be ``http://127.0.0.1:8889/tree`` and ``http://localhost:8889/tree`` . And also the name of the node your interactive session is running on. 
+
+Accessing Jupyter with a Web Browser
+-------------------------------------
+
+Once you have completed the above steps, you are ready to access Jupyter with a web browser. There are two ways to do this:
+
+#. Using flight-desktop (Recommended)
```
fredsonnenwald, line 76: I think there might be some assumptions here about your flight session being on the same node as your Jupyter session.

stanage/software/apps/jupyter.rst
```diff
@@ -0,0 +1,131 @@
...
+    
+    pip install jupyter
+
+5. If it's your first time using Jupyter, you will need to create a notebook configuration file. The following command will create ``jupyter_notebook_config.py`` in your home directory ``~/.jupyter/jupyter_notebook_config.py``.
+
+.. code-block:: bash
+    
+    jupyter notebook --generate-config
+
+6. Set up a password for your Jupyter notebook (this only needs doing once), with the following command:
```
fredsonnenwald, line 43: Really need to emphasise the need for there to be a password

hpc/sshkeys.rst
```diff
@@ -0,0 +1,97 @@
...
+    Enter passphrase (empty for no passphrase):
+    Enter same passphrase again:
+
+2. Once the keys have been succesfully created, we need to copy the public key to the HPC. Your public key is the file with the extension ``.pub``
+
+.. code-block:: bash
+
+    scp Path/To/File/id_rsa.pub <UserName>@<ClusterName>.sheffiled.ac.uk:~/.ssh/authorized_keys
+
+3. Add SSH private Key to mobaXterm or SSH-Agent on your local machine. This will vary based on your operating system and whether you are using a GUI ssh client like mobaxterm or ssh on a terminal.
```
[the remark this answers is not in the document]
jkwmoore: Personally I want to see a general move away from X11 forwarding and MobaXterm/Putty. The latter still leverage a separate exclusive SSH ecosystem which isolates users from the arguably much better OpenSSH stack and dependent software.

OpenSSH also natively supports the tooling for Smallstep (which enables increased security via MFA enabled ephemeral SSH key provisioning). Smallstep is likely to be an enforced posture for security / risk reasons when this becomes possible.

I would prefer to see us move to a posture where we divest from the Putty / Moba ecosystem in favor of native terminal tooling for each OS and/or leveraging OpenSSH which should be receiving regular vendor updates from the OS maintainer (other clients are getting less regular TLC in general I think.)


This gives the flexibility to use every client which supports the usage of SSH agents on any platform for both SSH and data transfer. Assuming we converge on a specific file transfer method (Filezilla?) this should also reduce support / training needs.

n.b. WSL SSH shouldn't be needed (but could be used) since Windows now has native OpenSSH support. I have requested that the managed desktop ensure this is also available everywhere.

stanage/software/apps/jupyter.rst
jkwmoore: I'm expecting the user should be making the request for the interactive flight-desktop session then spawning the notebook within it here.

hpc/sshkeys.rst
```diff
@@ -0,0 +1,97 @@
...
+This should produce a similar output to that shown below. You will be asked where the keys are to be stored and to enter a passphrase for your keys. **Do not create keys with out a passphrase**. If we detect a key without a passphrase a penalty may be added to your account.
+
+.. code-block:: bash
+
+    Generating public/private rsa key pair.
+    Enter file in which to save the key (/home/demo/.ssh/id_rsa):
+    Enter passphrase (empty for no passphrase):
+    Enter same passphrase again:
+
+2. Once the keys have been succesfully created, we need to copy the public key to the HPC. Your public key is the file with the extension ``.pub``
```
f-allian, line 31: @jkwmoore I think we should be more explicit here about the importance of the difference between these keys. Something like: "This will generate 2 different keys, 1 private one, and 1 public one. You ***must not!*** share your private key with anyone." etc.

stanage/software/apps/jupyter.rst
```diff
@@ -0,0 +1,131 @@
...
+==============
+
+.. sidebar:: |softwarename|
+
+   :URL: https://posit.co/products/open-source/rstudio/
+   :Documentation: https://jupyter-notebook.readthedocs.io/en/5.6.0/index.html
+
+
+Jupyter Notebook is an open source web application that you can use to create and share documents that contain live code, equations, visualisations, and text. 
+
```
f-allian, line 15: @jkwmoore With the conversations we had in Slack earlier this week, how much do we want to encourage using Jupyter notebooks for HPC purposes? Should we caveat this by saying something like "if you absolutely must use Jupyter Notebooks, then..."

stanage/software/apps/jupyter.rst
```diff
@@ -0,0 +1,131 @@
...
+This is mostly recommended for advanced users.
+
+1. You must have :ref:`ssh keys<ssh_keys>` in place to use this method.
+2. Open a new terminal  and enter the following command. (Please note , without SSH keys, this step will fail)
+
+.. code-block:: bash
+
+    ssh -J $YourUsername@stanage.sheffield.ac.uk -L 8888:127.0.0.1:8888 $YourUsername@<IP Address>
+
```
f-allian, line 106: @jkwmoore Might be worth highlighting this only works in Windows (i.e. powershell/cmd prompt) and not any other OS (from mine and Niki's testing).

f-allian (changes requested):

hpc/sshkeys.rst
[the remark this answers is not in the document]
fredsonnenwald: I really like the idea of using ssh built into Windows, particularly as for teaching it's an identical experience independent of OS. (And it's much easier than WSL!) 

I'm not sure built-in is necessarily better on a regular updates front. The file modification date for ssh on my Laptop is early 2022 with version OpenSSH_8.1p1. For context Ubuntu LTS desktop has 8.9p1, and 9.6p1 is the current release.

hpc/sshkeys.rst
```diff
@@ -0,0 +1,97 @@
...
+
+.. code-block:: bash
+
+    scp Path/To/File/id_rsa.pub <UserName>@<ClusterName>.sheffiled.ac.uk:~/.ssh/authorized_keys
+
+3. Add SSH private Key to mobaXterm or SSH-Agent on your local machine. This will vary based on your operating system and whether you are using a GUI ssh client like mobaxterm or ssh on a terminal.
+
++------------------------------------------------------+------------------------------------------------------+---------------------------------------------------------------------------------------------------+
```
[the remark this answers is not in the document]
nikishe: There are 2 interactive nodes, node 1 and node 2 . This table showed their IP address

hpc/sshkeys.rst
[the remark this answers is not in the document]
nikishe: @fredsonnenwald ssh is perfect and im currently planting the seed for tabby (supports all OSs) , but atm docs mostly  concentrate on mobaxterm as the recommended option but this is platform dependant and abit of a pain to support and teach as you have to give 3 different paths based on user OS .

hpc/sshkeys.rst
nikishe: thats me being inconsistent

stanage/software/apps/jupyter.rst
nikishe: I have gone back to the 4 users and we are having a stance on flight

nikishe:
After our discussion , I am closing this pull request. Might revive it once we get managed ssh keys

PR #2017, opened 2024-03-12
Add robots.txt to webroot.

jkwmoore:
As above. This should ensure other search engines find the sitemap.


jkwmoore:
Added you as reviewers so you're aware I've made this change. But going to merge now.

PR #2052, opened 2024-03-28
Replace env variables with generic paths

Carldkennedy:
Replacing with generic paths seem the best option here, taking into account all comments on the linked issue

### DIFF
--- a/hpc/transferring-files.rst
+++ b/hpc/transferring-files.rst
@@ -66,19 +66,19 @@ To upload, you transfer from your local machine to the remote cluster:
 
 .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


jkwmoore (approved):

jkwmoore (approved):
LGTM, but we may also want to amend any use of $VAR styling for things.

Carldkennedy:
@jkwmoore Suggest we replace $USER with YOUR_USERNAME throughout documentation except where $USER would be an appropriate bash expansion. And where possible use $HOME.

jkwmoore:
No objections to that, but if you were still considering mouse over pop ups this might be where they are useful. (i.e. pop up hint on what $USER actually is.)

PR #2062, opened 2024-04-02
Add FAQ on ACL change of ownership

nahceigna:
Hi all,

I have included an FAQ entry showing users how to change the ACL ownership of files/folders.


FAQs.rst
```diff
@@ -708,6 +708,45 @@ but cannot use more than 400 at once.
    **Total**                                            **52**
    ==============================================       =========================== 
 
-|br|
+How to change ownership of files/folders not as root?
+-----------------------------------------------------
+
+Normal users will not have the permission to **chown** since they are not being root or in sudoers. Therefore, the only valid method for transferring ownership of files/folders is to make
```
Carldkennedy, line 714: We try to be succinct where possible. Maybe something like:
Cluster users can not use **chown** for transferring ownership of files/folders, however it is possible using Access Contol Lists (ACLs).   
The continue with: The following assumes...

FAQs.rst
```diff
@@ -708,6 +708,45 @@ but cannot use more than 400 at once.
...
+the documents readable to the person who is taking ownership, reset the ownership, copy, delete and finally move the documents.
+
+This process will also be useful for users who end up with an older account and a newer account. It allows user to transfer data across local filesystems from older account to the newer.
+Please note that the below guide assumes that both accounts are still be active. If you have lost access to the old account in the last few weeks then get in touch with us via research-it@sheffield.ac.uk and we may be able to help transfer files across.
+
+The following assumes user1 is the current owner and user2 is going to be the new owner. Please change the ownership of the files/folders using the following instructions:
```
Carldkennedy, line 720:
```suggestion
The following assumes user1 is the current owner and user2 is going to be the new owner:
```

FAQs.rst
```diff
@@ -708,6 +708,45 @@ but cannot use more than 400 at once.
...
+
+1. user1 makes the files/folders to be changed ownership available to read by user2, most likely with Linux ACLs:
```
Carldkennedy, line 722:
```suggestion
1. user1 makes the files/folders available to read by user2 with Linux ACLs:
```

FAQs.rst
```diff
@@ -708,6 +708,45 @@ but cannot use more than 400 at once.
...
+.. code-block:: bash
+
+        cp -R /the/directory/changing/ownership my/tmp/directory
+
+3. user2 checks if they have the copy of the file with the correct ownership:
+
+.. code-block:: bash
+
+        ls -l
+
```
Carldkennedy, line 739: give an example output

FAQs.rst
```diff
@@ -708,6 +708,45 @@ but cannot use more than 400 at once.
...
+4. user1 deletes the existing folder recursively:
+
+.. code-block:: bash
+
+        rm -rf /the/directory/changing/ownership
+
+5. user2 moves the folder back where it was before:
```
Carldkennedy, line 746:
```suggestion
5. user2 moves the folder back to it's original location:
```

Carldkennedy (changes requested):
Well done on your first commit. Added a few comments and suggested a few changes.

FAQs.rst
```diff
@@ -708,6 +708,51 @@ but cannot use more than 400 at once.
    **Total**                                            **52**
    ==============================================       =========================== 
 
-|br|
+How to change ownership of files/folders not as root?
+-----------------------------------------------------
+
+Cluster users cannot use **chown** for transferring ownership of files/folders. However, it is possible using Access Control Lists (ACLs).
```
jkwmoore, line 714: For this part I think we will want to explain a little more background or users may have further unanswered questions. I was thinking about:

- The 'chown' command can only be successfully executed by the root user on any linux system(why? security - https://unix.stackexchange.com/questions/27350/why-cant-a-normal-user-chown-a-file/27374#27374)

- HPC users are not allowed to use the root account or assume its permissions (why? security on a shared system like HPC)

- Explain in human terms what we're doing before we give CLI type instructions e.g. we will bypass these limitations by giving the 2nd user read permissions on the data so they can make a copy of their own, then we can delete the original data.

jkwmoore (changes requested):
Where we have the usernames 'user1' and 'user2' in line, we might want to highlight these using bold formatting (`**bold**`) or via inline code highlighting (` ``inline code`` `).

If we're using FACLs we should probably also link out to some more detail and potentially provide a warning that FACLs can be hard to detect but can be seen via the extra ``+`` you can see via the ls command where they are present.

e.g.

```
my_user@work-ubuntu-vm:/tmp/tmp.x08Uc3hZJ1$ ls -al
total 28
drwx------   2 my_user my_user  4096 Apr  3 16:47 .
drwxrwxrwt  27 root   root   20480 Apr  3 17:12 ..
-rw-rwxr--+  1 root   root       0 Apr  3 16:47 testfile
my_user@work-ubuntu-vm:/tmp/tmp.x08Uc3hZJ1$ getfacl testfile 
# file: testfile
# owner: root
# group: root
user::rw-
user:my_user:rwx
group::r--
mask::rwx
other::r-
```

If we wanted to be gucci, we could also use this as an opportunity to try out ASCII cinema to provide a "running example".

https://asciinema.org/

https://pypi.org/project/sphinxcontrib.asciinema/

FAQs.rst
```diff
@@ -708,6 +708,72 @@ but cannot use more than 400 at once.
    **Total**                                            **52**
    ==============================================       =========================== 
 
-|br|
+How to change the ownership of files and folders when not the root user?
+------------------------------------------------------------------------
 
+Cluster users are not allowed to use the root account for system administrators to control and ensure the security of all users.
```
Carldkennedy, line 714:
```suggestion
Cluster users are not allowed to use the root account due to it's elevated privileges, it is reserved for system administrators to ensure the security of all users.
```

FAQs.rst
```diff
@@ -708,6 +708,72 @@ but cannot use more than 400 at once.
...
+Thus, they cannot use **chown** for transferring ownership of files/folders due to the possibility of leading to security vulnerabilities.
```
Carldkennedy, line 715:
```suggestion
Thus, they cannot use **chown** for transferring ownership of files/folders.
```

FAQs.rst
```diff
@@ -708,6 +708,72 @@ but cannot use more than 400 at once.
...
+However, it is possible using Access Control Lists (ACLs).
 
+The files/folders have to be stored in public Fastdata areas, detailed instructions are contained in the :ref:`fastdata_dir`.
```
Carldkennedy, line 718: Could have this as item 1 on the list. I.e make sure user2 has access to the folder/files

Carldkennedy (commented):
Added a few comments and suggestions

FAQs.rst
nikishe, line 714: You are not wrong but I think "HPC users" would more clear for newbies. Ill let you make the call though.

nikishe (commented):
Second pull request already. Good job. Few suggestion from the team , but this looks really good so far

FAQs.rst
jkwmoore: Here I think we probably want reword a bit to indicate 

> "Only system administrators are granted access to the root account (superuser privileges) for security reasons."

Can possibly link on `superuser privileges`  out to: https://en.wikipedia.org/wiki/Superuser

Side note on "it's" versus "its": the former contraction is equivalent to "it is", so should only be used where that works.

FAQs.rst
```diff
@@ -708,6 +708,72 @@ but cannot use more than 400 at once.
...
+        user:user2:r-x
+        group::r-x
+        mask::r-x
+        other::r-x
+
+4. **user2** copies the files from **user1**:
+
+.. code-block:: console
+
+        cp -R /mnt/parscratch/users/user1/public/the/directory/changing/ownership/ my/tmp/directory
```
beckywilks, line 763: If someone is following this through they probably won't have a tmp dir. Maybe add a line 'mkdir my/tmp/directory/'. I'd just make it part of step 2, no need for an extra step.

Most users will probably know to make a directory but I think it will make for a more streamlined process for users following it through. :)

FAQs.rst
jkwmoore: Think we're also missing some background on: "As using chown requires owning the file or being root for security reasons, thus it is not possible for a non-root user to reassign ownership."

Where "security reasons" can link out to https://unix.stackexchange.com/a/27374

nikishe (approved):
LGTM

beckywilks (approved):

Carldkennedy (approved):
LGTM

FAQs.rst
```diff
@@ -726,6 +726,80 @@ but cannot use more than 400 at once.
    **Total**                                            **52**
    ==============================================       =========================== 
 
-|br|
+How to change the ownership of files and folders when not the root user?
+------------------------------------------------------------------------
 
+Only system administrators are granted access to the root account (`superuser privileges <https://en.wikipedia.org/wiki/Superuser>`_) for security reasons.
+As using the **chown** command requires owning the file or being root for `security reasons <https://unix.stackexchange.com/a/27374>`_, it is not possible for a non-root user to reassign ownership.
+However, it is achievable using Access Control Lists (ACLs).
```
jkwmoore, line 734:
```suggestion
For security reasons only system administrators are granted access to the root account (`superuser privileges <https://en.wikipedia.org/wiki/Superuser>`_)
and as successfully using the **chown** command `requires root account permissions <https://unix.stackexchange.com/a/27374>`_ it is not possible for a non-root user to directly reassign ownership.
However it is possible to do so indirectly by using Access Control Lists (ACLs).
```

FAQs.rst
```diff
@@ -726,6 +726,80 @@ but cannot use more than 400 at once.
...
+
+7. **user2** checks if they have the copy of the files/folders with the correct ownership:
+
+.. code-block:: console
+
+        [user2@login1 [stanage] public]$ ls -l
+        total 4
+        drwxr-xr-x 2 user2 clusterusers 4096 Apr  3 14:00 the/directory/changing/ownership
+
+8. **user1** deletes the existing folder recursively:
```
jkwmoore, line 799: We might want a warning admonishment here that they should check with user2 the files have transferred prior to deletion as the fastdata areas do not have backups.

jkwmoore (approved):
LGTM - just a suggestion to making the wording a bit less clunky and we may want to add a warning pop out.

nikishe (approved):

jkwmoore (approved):
Added a missing separator but otherwise good to merge.

PR #2064, opened 2024-04-04
Add ref and commit to the footer.

jkwmoore:
This probably needs a bit of an update in the README to underline the dependency here on the theme directory - not just for this change but in general.

What this looks like:

image

With it linking out to the commit.

### DIFF
--- a/themes/sheffieldhpc/footer.html
+++ b/themes/sheffieldhpc/footer.html
@@ -1,0 +1,62 @@
+<footer>
+  {%- if (theme_prev_next_buttons_location == 'bottom' or theme_prev_next_buttons_location == 'both') and (next or prev) %}
+    {#- Translators: This is an ARIA section label for the footer section of the page. -#}
+    <div class="rst-footer-buttons" role="navigation" aria-label="{{ _('Footer') }}">
+      {%- if prev %}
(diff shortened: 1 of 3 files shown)


themes/sheffieldhpc/footer.html
```diff
@@ -0,0 +1,62 @@
...
+    {%- if build_id and build_url %}
+      <span class="build">
+        {#- Translators: Build is a noun, not a verb -#}
+        {%- trans %}Build{% endtrans -%}
+        <a href="{{ build_url }}">{{ build_id }}</a>.
+      </span>
+    {%- elif commit %}
+      <span class="commit">
+        {#- Translators: the phrase "revision" comes from Git, referring to a commit #}
+        {%- trans %}Revision: {% endtrans %} {%- if git_ref %}{{ git_ref }}{%- endif %} <a href="https://github.com/rcgsheffield/sheffield_hpc/commit/{{ commit }}">{{ commit }}</a>.
```
jkwmoore, line 36: This is the amended line we could make less brittle.

jkwmoore: e.g. with: `GITHUB_REPOSITORY` - https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

PR #2077, opened 2024-04-18
April 2024 newsletter: correction re whether access to Research Shared Areas always uses encryption in transit

willfurnass:


Carldkennedy (approved):
LGTM

PR #2080, opened 2024-04-19
Add Windows Terminal

Carldkennedy:
Recommend Windows Terminal on MD
Add instructions for adding profiles to Windows Terminal


jkwmoore (approved):
Added some minor changes to polish things up a tiny bit more. LGTM.

Before we merge, just want to check whether we'd prefer to use tabs to contain the profiles setup rather than reference info files?

Carldkennedy:
> Added some minor changes to polish things up a tiny bit more. LGTM.
> 
> Before we merge, just want to check whether we'd prefer to use tabs to contain the profiles setup rather than reference info files?

Having the profiles setups in the reference info section makes the connecting page easier to read and skimable. Do we agree?

PR #2085, opened 2024-04-23
Add GSL Stanage

Carldkennedy:
Adds GSL documentation for Stanage

### DIFF
--- a/stanage/software/libs/gsl.rst
+++ b/stanage/software/libs/gsl.rst
@@ -21,7 +21,7 @@ Usage
 
 .. note::
    
-   We only have Icelake builds at present. 
+   Currently, we only have builds which are compatible with our CPU nodes.
 
 The GSL library can be loaded by running one of: 
 
--- a/stanage/software/libs/gsl.rst
+++ b/stanage/software/libs/gsl.rst
@@ -16,12 +16,12 @@ The GNU Scientific Library (GSL) is a collection of routines for numerical compu
 The routines have been written from scratch in C.  
 See `here <https://www.gnu.org/software/gsl/doc/html/intro.html>`__ for the types of routines that the GSL provides.
 
-Usage
------
-
 .. note::
    
    Currently, we only have builds which are compatible with our CPU nodes.
+
+Usage
+-----
 
 The GSL library can be loaded by running one of: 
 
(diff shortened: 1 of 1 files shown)


stanage/software/libs/gsl.rst
```diff
@@ -0,0 +1,92 @@
...
+   :Documentation: https://www.gnu.org/software/gsl/doc/html/index.html
+
+The GNU Scientific Library (GSL) is a collection of routines for numerical computing. 
+The routines have been written from scratch in C.  
+See `here <https://www.gnu.org/software/gsl/doc/html/intro.html>`__ for the types of routines that the GSL provides.
+
+Usage
+-----
+
+The GSL library can be loaded by running one of: ::
```
willfurnass, line 22: Note that only have icelake builds at present?

willfurnass (approved):
LGTM.  See minor comment.

stanage/software/libs/gsl.rst
```diff
@@ -19,7 +19,13 @@ See `here <https://www.gnu.org/software/gsl/doc/html/intro.html>`__ for the type
 Usage
 -----
 
-The GSL library can be loaded by running one of: ::
+.. note::
+
```
willfurnass, line 23: Is this consistent with how we've handled this elsewhere?  should users know what 'Icelake' is before arriving at this page?

willfurnass: Perhaps we don't need to resolve this now, but I think it's worth thinking about what Research Support want the general approach to be on Stanage; I think at present there's the risk of confusion on GPU nodes regarding what's installed.

Carldkennedy: We could be indicate CPU and/or GPU node in the sidebar?

Using tabs, where available modules are different on each architecture?

willfurnass: @nikishe and @desmfryan Thoughts?

Carldkennedy: See #2087 for an example

nikishe (approved):

nikishe (approved):
Good work Carl

stanage/software/libs/gsl.rst
nikishe: Tabs to separate GPU/cpu seems good. You can have it as inner tabs as well as long as its not to squashed

Issue #1901, opened 2023-11-07
Poor connectivity guidance


Carldkennedy:
New section/ FAQ 
Tips and hints for dealing with poor connectivity issues whilst connected to the clusters.

jkwmoore:
Namely something along the lines of:

```
Host *.shef.ac.uk
  ServerAliveInterval 30
  ServerAliveCountMax 4
  ControlMaster auto
  ControlPath ~/.ssh/%r@%h:%p
  ControlPersist 600s
```

And it appears that the usage of Mosh should be possible in concert if provided on cluster (potentially as a module or via [Conda](https://anaconda.org/conda-forge/mosh).)

It is probably worth investigating and finding some kind of suitable implementation.

willfurnass:
Use of SSH multiplexing also suggested in Bede docs: https://bede-documentation.readthedocs.io/en/latest/faq/#how-do-i-reduce-the-number-of-times-i-m-prompted-for-my-password-and-a-mfa-code

jkwmoore:
We do have it in our docs, although it does seem a bit out of the way:  https://docs.hpc.shef.ac.uk/en/latest/hpc/mfa.html#ssh-clients

We might want to think about shifting this to be more visible too.

PR #2089, opened 2024-04-24
Poor connectivity

beckywilks:
FAQ for poor connectivity  #1901 

I took the information from [here](https://docs.hpc.shef.ac.uk/en/latest/hpc/mfa.html#tips-for-working-with-mfa) and made it into an FAQ so its easier to find/access. 

It seems MOSH would be difficult to implement and could potentially raise security problems so this has been ignored for now.

### DIFF
--- a/FAQs.rst
+++ b/FAQs.rst
@@ -726,6 +726,7 @@ but cannot use more than 400 at once.
    **Total**                                            **52**
    ==============================================       =========================== 
 
+-----
 
 How can I stay connected to the cluster for longer?
 ---------------------------------------------------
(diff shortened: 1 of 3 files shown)


jkwmoore (approved):
Added a missing separator but otherwise LGTM.

jkwmoore:
Just as a side comment (there's no issue with this here), where we use imports we sometimes have to take care with the TOCtree / header hierarchy (with the import's headers having to be subordinates of the page it is being imported into) which has caught me out before.

nahceigna (approved):

jkwmoore:
Feel free to give this one a merge. Probably better to avoid having to keep rebasing.

PR #2100, opened 2024-05-01
Addition of cuda 12.1.1 and cudnn 8.9.2 

nikishe:
Addition of cuda 12.1.1 and cudnn 8.9.2  to address some errors brough…t up with 12.0.0


Carldkennedy (approved):
LGTM

Issue #1963, opened 2024-01-25
Improvements to getting help page


Carldkennedy:
> Could something like this be implemented? Prompting users to give a clear description of their issue? `<a href="mailto:research-it@sheffield.ac.uk?body=Submitting a request for support:%0D%0AClear and concise description of the issue. Include specific details like job IDs, error messages, screenshots, and the steps leading to the problem.%0D%0AMention relevant system details such as which cluster, software/library versions, and any specific modules used.%0D%0AIf possible, include steps or scripts to reproduce the issue.">Email Support</a>`

I agree that guidance on _how_ to best ask for support would be nice (possibly formatted as bulleted list) but best addressed via a separate Issue/PR around improving the Getting Help page?

_Originally posted by @willfurnass in https://github.com/rcgsheffield/sheffield_hpc/issues/1903#issuecomment-1824363250_

PR #2106, opened 2024-05-07
Add email support template

nahceigna:
Hi all,

I added the email support template in [Need help?](https://docs.hpc.shef.ac.uk/en/latest/help.html#general-hpc-problems) section to provide a guidance on how to best ask for support.

Issue #1963

### DIFF
--- a/help.rst
+++ b/help.rst
@@ -40,6 +40,8 @@ Use the `Email support
 subject=[CLUSTER_USERNAME]%20Issue%20Summary&
 body=
 Hi%20there,%0D%0A%0D%0A
+I%20confirm%20that%20I%20have%20checked%20the%20
+FAQs%20at%20https://docs.hpc.shef.ac.uk/en/latest/FAQs.html#frequently-asked-questions%20for%20a%20possible%20solution%20to%20my%20issue.%0D%0A%0D%0A%0D%0A
 Issue%20Summary%0D%0A
 Please%20provide%20a%20clear%20and%20concise%20description%20of%20the%20issue.%0D%0A%0D%0A%0D%0A%0D%0A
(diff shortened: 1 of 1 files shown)


nahceigna:
> LGTM for the most part, might just want to:
> 
> 1. Extend the template a little to also request the job log files.
> 2. Have the user first confirm they've checked the FAQ page (provide this as a link).
> 3. Provide the template as a copy pasteable block as well.
> 
> Some folks may only use GMail and I don't know if clicking the link will open a GMail draft as we would desire.

@jkwmoore Thanks for your suggestions! Since the FAQ page is already referenced in the previous paragraph, do we need to confirm it with the users again in the paragraph discussing the email template?

jkwmoore (commented):
I was thinking of having it in the body of the email to essentially "trick" people into actually going and having a look. It's one of the kinda naughty psychological tricks to make people actually go do it as most people won't want to lie to us in their opening email for help.

nikishe:
Looks great. Only recommendation is probably to emphasize / encourage them to send scripts and logs as txt attachments rather than them  to paste them in the email; as this can get very messy if the odd person decided to paste 1000 lines of code/logs into the email.

nikishe (approved):
LGTM. Good work on the email template link.

jkwmoore (commented):
Not sure the email subject makes sense to me. Might be an idea to go with something like

Subject: "HPC Help request - [CLUSTER NAME] - [SHORT DESCRIPTION]"

Or something similar that will help support staff understand, assess effort and prioritise tickets at a glance. Besides that, looks good to merge.

nikishe (approved):

jkwmoore (approved):

PR #2122, opened 2024-06-01
R patch Stanage & Bessemer

desmfryan:
Added note in installation section that all versions patched for CVE vulnerability.

Updated modules available on Stanage.

Please have a quick scan & merge if all good.

Best wishes,

Des


nikishe (approved):
LGTM

PR #2142, opened 2024-06-26
Update Stanage libsndfile command

nahceigna:
Updated the command from:

```shell
gcc list_formats.c pkg-config --libs sndfile -o output
```

to the following:
```shell
gcc list_formats.c `pkg-config --libs sndfile` -o output
```


nikishe (approved):
Good catch

Carldkennedy (commented):
Apologies. This review was left in a unconfirmed state.

PR #2155, opened 2024-07-01
Add parallel EL9-beta build to site.

jkwmoore:
Amended workflow now builds a beta branch and also makes this available via ``en/el9-beta`` rather than ``en/latest`` assuming I've not fluffed anything.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -21,11 +21,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # Single deploy job since we're just deploying the once to Github Pages.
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-EL9-beta-branch.yml
```diff
@@ -0,0 +1,30 @@
+---
+name: test_docs_build_EL9_beta_branch
+on:
+  # Trigger on pull requests against the master branch
```
jkwmoore, line 4:
```suggestion
  # Trigger on pull requests against the EL9-beta branch
```

PR #2158, opened 2024-07-03
fftw and cfitsio EL9 docs

nikishe:
fftw and cfitsio docs for el9 upgrade


jkwmoore (approved):
LGTM. Please squash merge. A short PR description and more clear title is also not bad practise if we need to come back and look at things again later for some reason.

PR #2162, opened 2024-07-04
Git and git-lfs el9 docs

nikishe:
Had to introduce tabs due to the variations between gpu nodes and cpu nodes


Carldkennedy (approved):
LGTM. Made a few suggestions

PR #2169, opened 2024-07-04
Update GEOS EL9 doc

nahceigna:
Update versions to:

```bash
GEOS/3.9.1-GCC-10.2.0
GEOS/3.9.1-GCC-11.2.0
GEOS/3.10.3-GCC-11.3.0
GEOS/3.11.1-GCC-12.2.0
```


nikishe (approved):
LGTM

PR #2193, opened 2024-07-10
Cmake docs el9 cdk

### LINKED ISSUE #2132: CMake load GCC to fix linker issue - Stanage R9
Add note that if using CMake to compile then need to module load GCC/10.2 GCC/10.3 for those versions. Since new EL9 linker intentionally doesn't link  <= GCCcore/10.3

Example error:
```
usr/bin/ld.gold: /opt/apps/tuos/el9/software/live/GCCcore/10.2.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /usr/bin/ld.gold)
```

Fix:
```
if [[ "$version" == *GCCcore-10.2.0* ]]; then
    module load GCC/10.2.0
elif [[ "$version" == *GCCcore-10.3.0* ]]; then
    module load GCC/10.3.0
fi
```
https://uos-it-services.atlassian.net/browse/R9ST-79

Carldkennedy:
Updates versions for el9

Will close #2132

### DIFF
--- a/stanage/software/development/cmake.rst
+++ b/stanage/software/development/cmake.rst
@@ -16,21 +16,28 @@ CMake can be loaded with one of:
 
         .. code-block:: console
 
-           module load CMake/3.16.4-GCCcore-9.3.0   # compatible with foss-2021a toolchain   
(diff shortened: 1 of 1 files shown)


nikishe (approved):
Nice one on the note

PR #2204, opened 2024-07-16
Bump requirements and remove pathtools from requirements.

willfurnass:
Updated requirements are an amalgamation of several dependabot PRs.

pathtools is incompatible with Python 3.12 and doesn't seem to be needed in this project any more.


jkwmoore:
N.b. the separate EL9-beta branch will need a similar update.

